PR #2, opened 2023-04-26
Use FIELD_API to manage ecWam physics data fields

awnawab:
FIELD_API is a library used in the IFS physics to manage the movement of data fields between host and device. In this first implementation, an interface layer is created that allows FIELD_API to wrap the existing data allocations relevant to the ecWam physics. GPU enabled variants of ecWam can then simply use FIELD_API to manage device offload, rather than having to put offload directives in the code.

Eventually however, once the work of GPU adaptation for the ecWam physics and dynamics is in a more mature state, the derived-type definitions will be updated so that FIELD_API owns the data allocations, rather than wrapping existing allocations. This will enable the use of more performant FIELD_API features, for example, CUDA host-side pinned memory for much faster data transfers.


src/ecwam/yowfield_mod_config.yaml
```diff
@@ -0,0 +1,91 @@
...
+objtypes:
+ - frequency
+ - environment
+ - forcing_fields
+ - wave2ocean
+ - intgt_param_fields
+ - source_contribs
+
+objdef:
+  environment:
```
mlange05, line 21: The associated auto-generation of the data structures is very neat, but not obvious to someone who hasn't used fypp et al. before. From my reading of this, the re-generation is also optional, so maybe adding a short paragraph to the README about this option and how the data structures (wrappers) are generated could be useful?

src/ecwam/yowfield_mod.F90
```diff
@@ -0,0 +1,496 @@
...
+          REAL(KIND=JWRB), DIMENSION(:,:), POINTER :: CIWA=>NULL()
+          CLASS(FIELD_3D), POINTER :: F_WAVNUM=>NULL()
+          CLASS(FIELD_3D), POINTER :: F_CINV=>NULL()
+          CLASS(FIELD_3D), POINTER :: F_CGROUP=>NULL()
+          CLASS(FIELD_3D), POINTER :: F_XK2CG=>NULL()
+          CLASS(FIELD_3D), POINTER :: F_OMOSNH2KD=>NULL()
+          CLASS(FIELD_3D), POINTER :: F_STOKFAC=>NULL()
+          CLASS(FIELD_3D), POINTER :: F_CIWA=>NULL()
+      CONTAINS
+         PROCEDURE :: INIT => FREQUENCY_FIELD_INIT
```
mlange05, line 33: It looks like there is no finalisation for these calls that deallocated the FIELD objects allocated in INIT? If that is true, there is likely a small memory leak happening after each time step, which can easily be fixed by adding the corresponding destructor.

mlange05 (changes requested):
Ok, first of all I think this is great work with regards to enabling further GPU adaptation efforts, and a very nice and clean implementation of the minimally intrusive "wrapper" functionalities that the FIELD API provides. I've flagged two small issues, one a potential memory leak and one a small request for documentation. Once addressed I'm happy to approve this from a technical perspective.

However, mind you, I've only done a technical review of the PR and implementation itself, so the decision whether to adopt this long-term for ecWAM should be approved by one of the core maintainers (@wdeconinck, Jean Bidlot?). 

I'd also like to point out that this current implementation only introduces the "wrapper" mode of FIELD API objects, which are meant to be backward compatible with existing IFS-style memory allocations. In future efforts (once this is adopted), it would also be possible to further adap ecWAM to use these FIELD objects as the sole owner of core arrays, where the I/O routines access the arrays via FIELD objects and we can persist them across time steps (at least in the standalone). This would potentially reduce the code overhead in the main timestep loop and remove the need to run-time wrap existing arrays, but clearly this is potential future work.

src/ecwam/yowfield_mod_config.yaml
awnawab: Yes fully agreed, that would be a great addition! I'll add one 👍

src/ecwam/yowfield_mod.F90
awnawab: Ah very good catch! I'll add destructor calls, thanks!

src/ecwam/wavemdl.F90
wdeconinck, line 1: Could we attempt an implementation where TIME1 is not part of the signature of WAVEMDL as this is the outward facing routine, and is being called from the IFS
Could the same information not be accessed and/or exposed via GSTATS perhaps?

share/ecwam/scripts/ecwam_run_model.sh
```diff
@@ -284,7 +284,7 @@ echo "\n\n\t Running ${LAUNCH_PARALLEL} $(which ${MODEL}) took $DIFF seconds\n"
 
 trace_ls $(pwd)
 
-if [ "$(read_config validation.${precision}_precision) --default=NOTFOUND" != NOTFOUND ]; then
+if [ "$(read_config validation.${precision}_precision --default=NOTFOUND)" != NOTFOUND ]; then
```
wdeconinck, line 287: Good catch!

wdeconinck (changes requested):
I think it is a great addition! Thanks!

mlange05 (approved):
Great, I think everything has been addressed, and looks very neat. Great work, GTG from me. :shipit:

jrbidlot:
Hi Ahmad,  sorry it took so long to have a look at your tremendous piece of work.
Thank you for cleaning up the code as you went along to implement the new wrappers.
I will not pretend to understand fully how they work but once in place it seems not to difficult to modify when new fields might be needed.
When will these changes be reintegrated with 49R1. I have made some code changes that are now part of for_CY49R1 v11 but I don't think there might be many conflicts. Josh on the other hand may have started working on adding extra functionality. He might have introduced new field(s). So we might need to bring him up to speed with your work.

awnawab:
Hi Jean! Thanks a lot for taking the time to review the PR 🙏  These changes should go into CY49R1 relatively soon, because I have a deeper FIELD_API integration planned and it's best to apply these changes to ifs-source in two stages, starting with this PR. As for your remark about adding fields, yes it should be quite straightforward to add fields by updating the yaml file (as long as we are not introducing nested derived-types). However if Josh should have any questions, my office is next door to his and I would be more than happy to sit down with him and go through the changes I've made.

wdeconinck:
Thank you all for the review. Great work. I am merging this now.

PR #3, opened 2023-08-09
Add ecwam-bundle to main repo

awnawab:
This PR adds the ecwam-bundle. A few arch files, for Intel builds on the ATOS and GNU builds on the macs, are also added. The `README` has been updated to reflect the inclusion of the bundle.

NB: I have left the github actions config untouched because I am not that proficient at modifying it. Please feel free to add commits to this branch as necessary to use the bundle in the CI builds.


README.md
```diff
@@ -74,30 +71,42 @@ If you want to pre-download or install extra data files, run in the source-direc
 
     $ share/ecwam/data/populate.sh
 
-You must compile ecWAM out-of-source, so create a build-directory
+The preferred method for building ecWAM is to use the bundle definition shipped in the main repository. For this please build the bundle via:
```
reuterbal, line 74: I'm not sure we want to advertise this at the preferred method but rather provide it as a comfortable one-stop solution. Documenting the standalone build as the default is probably better imho.

awnawab: Hmm yes, thinking back on it I do agree with you. I'll update the README accordingly. Thanks for the suggestion 😄

arch/ecmwf/hpc2020/default/env.sh
```diff
@@ -0,0 +1,48 @@
...
+[[ ${IFS_RUNTIME_ENV:-unset} == "unset" ]] && module_purge
+
+# Load modules
+module_load prgenv/intel
+module_load intel/2021.4.0
+module_load hpcx-openmpi/2.9.0
+module_load intel-mkl/19.0.5
+
+# Don't load these modules if env.sh is used as part of the IFS runtime environment - only the modules above are required
+if [[ ${IFS_RUNTIME_ENV:-unset} == "unset" ]]; then
```
wdeconinck, line 29: This is not to be used via IFS so no need for this check.

arch/ecmwf/hpc2020/intel/2021.4.0/hpcx-openmpi/2.9.0/env.sh
```diff
@@ -0,0 +1,48 @@
...
+  echo "+ module unload $*"
+  module unload $*
+}
+module_purge() {
+  echo "+ module purge"
+  module purge
+}
+
+# Unload all modules to be certain
+[[ ${IFS_RUNTIME_ENV:-unset} == "unset" ]] && module_purge
```
wdeconinck, line 20: no need for the IFS check

arch/ecmwf/hpc2020/intel/2021.4.0/intel-mpi/2021.4.0/env.sh
```diff
@@ -0,0 +1,49 @@
...
+  echo "+ module unload $*"
+  module unload $*
+}
+module_purge() {
+  echo "+ module purge"
+  module purge
+}
+
+# Unload all modules to be certain
+[[ ${IFS_RUNTIME_ENV:-unset} == "unset" ]] && module_purge
```
wdeconinck, line 20: no need for IFS check

arch/ecmwf/hpc2020/intel/2021.4.0/intel-mpi/2021.4.0/env.sh
```diff
@@ -0,0 +1,49 @@
...
+
+# Load modules
+
+module_load prgenv/intel
+module_load intel/2021.4.0
+module_load intel-mpi/2021.4.0
+module_load intel-mkl/19.0.5
+
+# Don't load these modules if env.sh is used as part of the IFS runtime environment - only the modules above are required
+if [[ ${IFS_RUNTIME_ENV:-unset} == "unset" ]]; then
```
wdeconinck, line 30: no need for IFS check

bundle.yml
```diff
@@ -0,0 +1,53 @@
...
+        git     : https://github.com/ecmwf/eccodes
+        version : master
+        cmake   : ENABLE_MEMFS=ON
+
+    - fiat :
+        git     : https://github.com/ecmwf-ifs/fiat 
+        version : main 
+
+    - field_api :
+        git     : ${BITBUCKET}/RDX/field_api
```
wdeconinck, line 27: Is this publicly accessible?

wdeconinck (commented):
I would remove the apple arch
People should just use their system setup
and use brew to install packages.
It is hard to cater for a user's specific setup

wdeconinck:
I would also create a "package" subdirectory where these new files could be added.
We could envision also other packaging mechanisms to be added (spack, easybuild) ?
Also not to pollute the root dir too much with ecbundle specific stuff.

bundle.yml
awnawab: No it isn't. Because it's optional, the bundle create step will complete successfully with a warning (just tested on my laptop). This is the same setup as the CLOUDSC bundle. field_api will be hosted on github and open-sourced soon (~2 months); I'll update this once that happens.

awnawab:
Thanks a lot for the review and comments @wdeconinck! I've implemented all of your suggestions, please take another look when you get the chance 😄

wdeconinck (approved):

.gitignore
```diff
@@ -8,7 +8,9 @@ doc/latex
 .nfs*
 build/*
 install/*
-env.sh
+source/*
+ecbundle
+/env.sh
```
wdeconinck, line 13: Will this still work with change to package/bundle ?

awnawab: Yes, unless you put a backslash in front of it, the entries in `.gitignore` are applied recursively.

PR #1, opened 2023-04-17
Changed derived-type data structures to contiguous SoA layout

awnawab:
All the derived-type data structures are now stored in a contiguous structure-of-arrays (SoA) layout, as opposed to an array-of-structures (AoS) layout. The results are bit-identical, and there is a small uplift in performance. This was part of the preparatory work done for GPU porting.

PR #4, opened 2023-08-28
Various fixes for single-precision and switch to new FIELD_API

awnawab:
This PR adds various fixes required to validate successfully in single-precision for Intel and nvhpc builds. Due to the need to run NEMO in double-precision even if ecWAM is single-precision, a mixed precision FIELD_API build was needed. Rather than adding this functionality to the old (private) FIELD_API on bitbucket, ecWAM has been switched over to the new FIELD_API which already supports mixed-precision builds. As this repo will soon be open sourced (see [PR #1](https://github.com/ecmwf-ifs/field_api/pull/1)), I have made FIELD_API a hard dependency of ecWAM.


cmake/ecwam_fetchcontent_field_api.cmake
```diff
@@ -0,0 +1,34 @@
...
+set(clone_field_api TRUE)
+
+## Test if field_api was cloned by ecWAM
+if(field_api_FOUND)
+   cmake_path(RELATIVE_PATH field_api_DIR BASE_DIRECTORY ${CMAKE_BINARY_DIR} OUTPUT_VARIABLE path_var)
+   cmake_path(GET path_var PARENT_PATH parent_path)
+   string(FIND ${parent_path} "../" result_var)
+
+#  If field_api is found but was not cloned by ecWAM, clone_field_api is set to FALSE
+   string(COMPARE EQUAL ${result_var} "-1" clone_field_api)
```
awnawab, line 20: Once we build ecWAM and run `FetchContent` on FIELD_API, the cache variable `field_api_FOUND` is set to TRUE. If we then run `make clean` and re-configure the build without deleting the CMake cache, `field_api_FOUND` will still be true but the FIELD_API builddir will be empty and ecWAM will fail at compile-time when trying to use `libfield_api`.

If on the other hand `field_api_FOUND` is TRUE not because ecWAM cloned FIELD_API but because it is coming from a bundle or system installation, then the above mechanism ensures we don't run `FetchContent` on FIELD_API.

cmake/ecwam_fetchcontent_field_api.cmake
```diff
@@ -0,0 +1,34 @@
...
+endif()
+
+if( ${clone_field_api} )
+   include(FetchContent)
+   FetchContent_Declare(
+      field_api
+      GIT_REPOSITORY git@github.com:ecmwf-ifs/field_api.git
+      GIT_TAG naan-ecbuild
```
awnawab, line 28: This will be replaced by a tag once the new FIELD_API is open-sourced and [PR #2](https://github.com/ecmwf-ifs/field_api/pull/2) is approved. I will leave it as a draft-PR until then.

package/bundle/bundle.yml
```diff
@@ -23,22 +23,17 @@ projects :
...
-           ENABLE_FIELD_API_TESTS=OFF
-
     - ecwam :
         dir     : $PWD
         version : main
         require : fiat 
 
 options :
 
+    - single-precision :
```
wdeconinck, line 33:
```suggestion
    - with-single-precision :
```

cmake/ecwam_fetchcontent_field_api.cmake
```diff
@@ -0,0 +1,34 @@
...
+if(field_api_FOUND)
+   cmake_path(RELATIVE_PATH field_api_DIR BASE_DIRECTORY ${CMAKE_BINARY_DIR} OUTPUT_VARIABLE path_var)
+   cmake_path(GET path_var PARENT_PATH parent_path)
+   string(FIND ${parent_path} "../" result_var)
+
+#  If field_api is found but was not cloned by ecWAM, clone_field_api is set to FALSE
+   string(COMPARE EQUAL ${result_var} "-1" clone_field_api)
+endif()
+
+if( ${clone_field_api} )
```
wdeconinck, line 23:
```suggestion
if( clone_field_api )
```

wdeconinck (commented):
As far as the CMake configuration is concerned I am happy with this.

The DEL1 change means a non-bit reproducible change. This probably invalidates existing hashes in tests.

cmake/ecwam_fetchcontent_field_api.cmake
```diff
@@ -0,0 +1,34 @@
...
+   include(FetchContent)
+   FetchContent_Declare(
+      field_api
+      GIT_REPOSITORY git@github.com:ecmwf-ifs/field_api.git
+      GIT_TAG naan-ecbuild
+      OVERRIDE_FIND_PACKAGE
```
lukasm91, line 29: This requires CMake version 3.24 i.e. `cmake_minimum_required(VERSION 3.24)`. I guess this is fine, but I strongly recommend to adapt the CMake file. The error you get with 3.22 is extremely unclear.

awnawab: Thanks a lot for pointing this out! FIELD_API itself requires CMake 3.25 for fixes to the cmake `find_package(OpenaCC)`, and given that this PR makes ecWAM hard reliant on FIELD_API, it makes sense to update the minimum CMake version in ecWAM to 3.25 too.

reuterbal: If the dependency on 3.25 is only due to the OpenACC fixes, I'd argue it would be better to simply add the following workaround and make that a less strict requirement:

https://github.com/ecmwf-ifs/ectrans/blob/e7a738749fa9f6e28cc1ecce35f1643700d2bb19/CMakeLists.txt#L38-L47

awnawab: `OVERRIDE_FIND_PACKAGE` already needs CMake 3.24 as Lukas pointed out. When reconfiguring after a `make clean`, `OVERRIDE_FIND_PACKAGE` makes the `FetchContent` mechanism a bit more robust. So the decision would be between 3.24 and 3.25. In that context, I think 3.25 isn't too restrictive. What are your thoughts on this @wdeconinck ?

wdeconinck: I think 3.24 is my current preference, using @reuterbal 's suggestion,  as Leonardo currently uses cmake 3.24 as latest version. Both LUMI and our own HPC have 3.25 however.

awnawab:
Hi @wdeconinck. Now that FIELD_API has been open-sourced, this is ready for review. I've removed the `+DEL1` to prevent divide by zero, so this PR won't affect bit-identicality. I get around the FPE by disabling divz checks for that file, and the run still validates correctly. I know it's a bit of a hack, but breaking bit-identicality just for NVHPC single-precision debug builds seems like overkill.

cmake/ecwam_fetchcontent_field_api.cmake
```diff
@@ -0,0 +1,35 @@
...
+#  If field_api is found but was not cloned by ecWAM, clone_field_api is set to FALSE
+   string(COMPARE EQUAL ${result_var} "-1" clone_field_api)
+endif()
+
+if( clone_field_api )
+   include(FetchContent)
+   FetchContent_Declare(
+      field_api
+      GIT_REPOSITORY https://github.com/ecmwf-ifs/field_api.git
+      GIT_TAG v0.2.1
```
mlange05, line 28: [no action] Eventually we might want a mechanism to define this in the top-level CMakeLists.txt, but not needed for this PR, I think

mlange05 (approved):
The FIELD API data structure changes look good to me and are consistent with the recent IFS update and the proposed release `v0.2.1` (sync'd with Meteo-France recently). The cmake-fetch_content mechanism also looks sound to me, although wiser heads can probably judge this better. :wink: 

GTG from me, and very nicely done @awnawab :tada: :smile: :+1: :pray:

wdeconinck (approved):
This addressed all my previous concerns and recommendations.
Very nice work @awnawab . Thank you!
I have merged your other PR.
You can rebase and merge when you're ready.

wdeconinck:
All green now :) You can merge when you're ready.

PR #10, opened 2024-02-09
Fix for building with single-precision NEMO

awnawab:
The FIELD_API utility functions were not compatible with single-precision NEMO builds. This was exposed during the testing for CY49R2. This PR is a replication of the fix filed in ifs-source.


wdeconinck:
Hi @awnawab I just noticed that with this PR we have two different preprocessor definitions now to control precision of ocean model: `PARKIND1_SINGLE_NEMO` and `WAM_NEMO_SINGLE`.
Perhaps this can be consolidated?

awnawab:
Hi @wdeconinck. Yes that makes a lot of sense, will do. Thanks!

wdeconinck (approved):

PR #12, opened 2024-03-21
 Compatibility with Loki v0.2

reuterbal:
This will make ecWAM processing compatible with the new Scheduler in Loki v0.2. Once https://github.com/ecmwf-ifs/loki/pull/213 is merged, we need this change to make it compatible with the new Loki version.


awnawab (approved):
Thanks a lot for testing with the new scheduler and implementing these fixes 🙏  I'm preparing a PR to main for the offloaded physics, and I'll use these to switch that over to the new scheduler once the Loki PR is merged.

PR #14, opened 2024-04-05
Ocean model precision hotfix

awnawab:
The ocean model precision checks should come after `HAVE_SINGLE_PRECISION` has been initialised. Could you please tag a hotfix release once merged? Thanks!


wdeconinck (approved):
Very sensible, thanks!

PR #15, opened 2024-04-17
GPU offload of physics kernel using LOKI

awnawab:
This is the second of three PRs aimed at merging the GPU offload capabilities to main.

The main focus of this PR is the CMake setup needed for Loki transformations, as well as changes to the FIELD_API glue code to facilitate offload.

The test configuration has also been updated to include a new test for the updated physics config in cy49.

One notable omission from this PR are unit-tests for the Loki transformations. I would be happy to include them, but I think the more appropriate place to test these would be the Loki regression tests. Once this PR is merged, I plan on updating the Loki regression tests to run off of the develop branch rather than naan-phys-gpu.


CMakeLists.txt
```diff
@@ -107,6 +108,28 @@ if( HAVE_OMP )
...
 
+### Apply Loki source-to-source transformations to the ecWam physics
+ecbuild_add_option( FEATURE WAM_LOKI
+                    DEFAULT OFF
+                    DESCRIPTION "Apply Loki transformations to ecWam physics"
+                    CONDITION loki_FOUND )
+
+# Determine Loki transformation type
+if( NOT DEFINED LOKI_MODE )
+   set( LOKI_MODE "scc-hoist" CACHE STRING "Transformation mode for Loki source transformations" )
```
mlange05, line 119: Do we want "scc-hoist" as default or "scc-stack". Are they performance equivalent?

package/bundle/bundle.yml
```diff
@@ -29,11 +30,18 @@ projects :
         version : v0.2.2
         cmake   : >
             ENABLE_SINGLE_PRECISION=OFF
-            ENABLE_ACC=OFF 
+
+    - loki :
+        git     : https://github.com/ecmwf-ifs/loki
+        version : main
```
mlange05, line 36: For release purposes we might want to use a tag here once the latest main is tagged (very soon). Locally developing and testing would continue on main, but for releases we want something non-moving, I'd say.

src/ecwam/yowfield_mod.fypp
```diff
@@ -53,7 +60,27 @@ MODULE YOWFIELD_MOD
...
+      #:for var, type, dim in objdef[obj]
+          #{if type == 'real'}#REAL(KIND=JWRB)#{elif type == 'ocean'}#REAL(KIND=JWRO)#{else}#INTEGER(KIND=JWIM)#{endif}#, DIMENSION(${','.join(':' for _ in range(dim))}$), INTENT(OUT), POINTER, CONTIGUOUS, OPTIONAL :: ${var.upper()}$
+      #:endfor
 
+      #:for var, type, dim in objdef[obj]
+          IF(PRESENT(${var.upper()}$)) CALL SELF%F_${var.upper()}$%GET_DEVICE_DATA_RDWR(${var.upper()}$)
+      #:endfor
+      END SUBROUTINE ${obj.upper()}$_UPDATE_DEVICE
+
+      SUBROUTINE ${obj.upper()}$_ENSURE_HOST(SELF)
```
mlange05, line 76: Minor nitpick: I'd recommend staying close to the FIELD API here and change the name to `<>_SYNC_HOST(SELF)`.

mlange05 (approved):
Ok, this looks very, (very!) good to me. I cannot speak to the test changes or the local source changes, only that set of source changes looks reasonable to me. Config and cmake setup look fine and clean to me too. 

I've left a few minor nitpicks or comments that could be fixed before merging, but none of them are showstoppers for me. Very nice work, @awnawab, good to go for me. 👍 

Final note, I've not actually run the offline version myself, but will try soon.

CMakeLists.txt
awnawab: scc-hoist is marginally faster than scc-stack. The main reason I set scc-hoist as the default was that scc-stack will need a small tweak to be performant in single-precision because it currently pads all allocations to multiples of 8 bytes.

package/bundle/bundle.yml
awnawab: Yes good catch! I'll update this to a tag once Loki main is tagged. Thanks!

src/ecwam/yowfield_mod.fypp
awnawab: Well spotted, this is a hangover from the old FIELD_API. I'll fix it, thanks!

CMakeLists.txt
mlange05: Ok, that's cool. Let's leave `scc-hoist` as default then.

CMakeLists.txt
```diff
@@ -107,6 +108,28 @@ if( HAVE_OMP )
   set( OpenMP_Fortran_LIBRARIES OpenMP::OpenMP_Fortran )
 endif()
 
+### Apply Loki source-to-source transformations to the ecWam physics
+ecbuild_add_option( FEATURE WAM_LOKI
```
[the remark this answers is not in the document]
awnawab: Agreed, I'll change the name of the option to `ECWAM_LOKI` 👍

CMakeLists.txt
```diff
@@ -107,6 +108,28 @@ if( HAVE_OMP )
...
+                    DEFAULT OFF
+                    DESCRIPTION "Apply Loki transformations to ecWam physics"
+                    CONDITION loki_FOUND )
```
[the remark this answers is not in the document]
awnawab: Well spotted, fixed!

CMakeLists.txt
```diff
@@ -107,6 +108,28 @@ if( HAVE_OMP )
...
+if( NOT DEFINED LOKI_MODE )
+   set( LOKI_MODE "scc-hoist" CACHE STRING "Transformation mode for Loki source transformations" )
+endif()
+
+# Workaround for the fact that CMake only sets language components to FOUND,
+# so OpenACC_FOUND must be set manually
+find_package( OpenACC )
+if( OpenACC_Fortran_FOUND AND OpenACC_C_FOUND )
+  set( OpenACC_FOUND ON )
+endif()
```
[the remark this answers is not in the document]
awnawab: Very good point, without making this logic explicit I will definitely forgot to drop this workaround in the future.

src/ecwam/CMakeLists.txt
```diff
@@ -356,6 +357,11 @@ list( APPEND ecwam_srcs
   z0wave.F90
 )
 
+if(HAVE_WAM_LOKI AND NOT LOKI_MODE MATCHES "idem|idem-stack")
+  list(APPEND ecwam_srcs wamintgr_loki_gpu.F90)
+  list(REMOVE_ITEM ecwam_srcs wamintgr.F90)
+endif()
```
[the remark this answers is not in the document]
awnawab: Yes it's definitely not obvious why this is needed so a comment would indeed be good 😅

CMakeLists.txt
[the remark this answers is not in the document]
wdeconinck: I'd say just feature `LOKI`.
In a bundle you could still control this with "${PNAME}_ENABLE_LOKI" if you need more fine-grained control.

awnawab:
Many thanks again all for your feedback! I have addressed all of your suggestions. Please have another look whenever you can 😄

.github/workflows/build.yml
```diff
@@ -115,6 +119,12 @@ jobs:
           brew install libomp
           brew install libaec
           brew install coreutils
+          brew install pyenv
+
+          pyenv install 3.11.4
+          pyenv global 3.11.4
+
+          echo "LOKI_PYTHON_ROOT_DIR=$HOME/.pyenv/versions/3.11.4/bin" >> $GITHUB_ENV
```
reuterbal, line 127: Any particular reason for using pyenv instead of the Github Action to install a specific Python version?
https://github.com/marketplace/actions/setup-python

awnawab: There is already a dependency on python3.12 indirectly via libomp (libomp => lit => python3.12). So rather than getting into the complexity of pinning down versions of the dependencies so we don't need a python newer than 3.11, I thought it was safer and less fragile to just have a sandboxed python3.11 installation that doesn't live on the path and is just used for loki.

reuterbal: Ah, ok, hadn't realized that. Makes sense to work around this with pyenv

.github/workflows/build.yml
```diff
@@ -47,6 +47,7 @@ jobs:
             compiler_cxx: g++-10
             compiler_fc: gfortran-10
             caching: true
+            cmake_options: "-DECBUILD_Fortran_FLAGS=-fcray-pointer"
```
wdeconinck, line 50: Why do we need "-fcray-pointer" ?
If this is a requirement for e.g. gfortran, then it should be in the ecwam cmake repository instead.

tests/etopo1_oper_an_fc_O48_iphys_0.yml
```diff
@@ -98,23 +97,23 @@ validation:
...
       relative_tolerance: 1.e-6
-      hashes: ['0x3F92366000000000']
+      hashes: ['0x3F978C93C0000000']
     - name: swh
       time: 2023-01-01 06:00:00
-      maximum: 0.7659064292907715E+01
+      maximum: 0.7301169872283936E+01
       relative_tolerance: 1.e-6
-      hashes: ['0x401EA2E1C0000000']
+      hashes: ['0x401D3465E0000000']
```
wdeconinck, line 119: Not bit-identical?

.github/workflows/build.yml
awnawab: We only need this for the loki idem-stack transformation, which I am running in the CI just to test the loki setup. You shouldn't normally run this configuration for CPU builds so I didn't want to include it in the compile flags macro.

tests/etopo1_oper_an_fc_O48_iphys_0.yml
awnawab: Previously I had lumped the `iphys=0` option with `llgcbz0=true`. But then I learned from Jean that in cycle 49, we run with `iphys=1` and `llgcbz0=true`. So I changed the `iphys=0` test to run with `llgcbz0=false` and added a new test with `iphys=1` and `llgcbz0=true`.

awnawab:
Hi @wdeconinck. As discussed offline, I've moved the `-fcray-pointer` flag to `ecwam_compiler_flags.cmake`.

wdeconinck (approved):

package/bundle/arch/ecmwf/hpc2020/nvhpc/22.11/toolchain.cmake
```diff
@@ -14,6 +14,20 @@
 # unless its stored as a cache variable
 set( OpenMP_Fortran_FLAGS "-mp=bind,allcores,numa" CACHE STRING "" FORCE)
 
+####################################################################
+# OpenAcc FLAGS
+####################################################################
+
+set( OpenACC_Fortran_FLAGS "-acc=gpu -gpu=cc80,lineinfo,fastmath" CACHE STRING "" )
```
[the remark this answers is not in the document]
wdeconinck: @awnawab , just pointing out this unresolved comment.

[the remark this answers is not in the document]
awnawab: Unless I was mistaken, I think comment is what @reuterbal was referring to here:
https://github.com/ecmwf-ifs/ecwam/pull/15/files#r1575882775

PR #17, opened 2024-05-06
GPU offload for wave propagation kernel

awnawab:
This is the last of the three PRs needed to add GPU acceleration to the wave model timestep.

The work included herein is entirely that of @fdisante and his colleagues at CINECA. Many thanks again for your efforts 🙏 

GPU optimised code for the source term integration scheme is generated by Loki at build time. This was possible to do because that kernel follows the single column computation pattern used throughout the IFS, and for which we have Loki transformation recipes.

The wave propagation kernel however has very different and bespoke memory access patterns, for which we do not have Loki recipes. Moreover, the wave propagation kernel is not expected to see as frequent scientific development as the source term integration scheme, so there is also less of a justification to develop Loki recipes for it.

That's why for the wave propagation kernel we have opted for a manual GPU port. This is just OpenACC for now, but OpenMP offload support will also be added at a later date. If we feel the need in the future that the wave propagation kernel would also benefit from auto-generation of GPU optimised code at build-time, then we can develop Loki recipes for that too.

Finally, all the changes herein have been tested in coupled IFS+nemo runs and are confirmed bit-identical.

PR #18, opened 2024-05-13
IFS-3459 post-processing output fix

awnawab:
This PR ports over the recent post-processing output fix filed by @jrbidlot. Ref: https://jira.ecmwf.int/browse/IFS-3459?filter=-1

I have also tested this in coupled tco399 runs with ifs-source develop.


wdeconinck (approved):
Rebased on develop after merging #17 for better code review.
Looks good to me! @jrbidlot could you cast one last look as well?

jrbidlot (approved):
It's good to go.

PR #21, opened 2024-06-12
Make fypp and pyyaml hard dependencies

awnawab:
This PR makes fypp and pyyaml hard dependencies, and therefore updates ecWAM to use the yaml reader provided in fckit 0.13.0.

Default optimisation flags are now also set explicitly rather than relying on the compiler default for optimised builds. This change has been tested as bit-identical in a coupled forecast in both SP,DP.

There is also a fix to how GPU specific optimisation flags are set to enable ecWAM offload to be integrated with ecPhys offload.

Once approved and merged, could this please be tagged as version 1.2.1?


wdeconinck (approved):

awnawab:
Hi @wdeconinck. Sorry for letting this hang for so long. I've now rebased on develop. Please review/merge at your convenience. Once merged, could you please tag this as 1.3.1? Many thanks again 🙏

PR #22, opened 2024-06-14
Incorporate WAM changes from IFS CY49R1

cxzk:
This incorporates the changes that were done in ecWAM in IFS CY49R1 (where it's still integrated into the IFS source) into the standalone version of ecWAM.

* Contains the ecWAM part of [IFS-3352](https://git.ecmwf.int/plugins/servlet/jira-integration/issues/IFS-3352)


src/ecwam/get_preset_wgrib_template.F90
```diff
@@ -25,7 +25,10 @@ SUBROUTINE GET_PRESET_WGRIB_TEMPLATE(CT,IGRIB_HANDLE)
 !     SUBROUTINE GET_PRESET_WGRIB_TEMPLATE(CT, IGRIB_HANDLE)
 !                INPUT:
 !                CT           : "I" for INTEGRATED PARAMETERS AND
+!                               "2" for INTGRATED PARAMETERS in GRIB 2                            
 !                               "S" for SPECTRA
+!                OPTIONAL INPUT:
+!                NGRIBV         GRIB VERSION TO BE USED (if absent then = NGRIB_VERSION)
```
wdeconinck, line 31: Is there some kind of redundancy with `CT=2` and `CT=I  AND NGRIBV=2` ?
Why is the "CT=2" needed if one can add the NGRIBV.
If this is redundant, I think the case `CT=2` should be removed, and the new `NGRIBV=2` option should be used.

Also there is a typo in the comment for "2" -->  "INTGRATED"

cxzk: I have no idea, I'm afraid. All I've done is cherry-pick the commits that have already been accepted into IFS CY49R1.

wdeconinck: @jrbidlot could you advise here please?

wdeconinck:
Merging this as-is, and then this can be followed-up later, if needed at all.

src/ecwam/get_preset_wgrib_template.F90
jrbidlot: Willem is correct. Actually, it is the comment line (line 28) that is incorrect because I mist have realised that indeed what Willem noted should be done and modified the code accordingly except for not removing line 28.

jrbidlot (commented):
It seems to correspond my last set of changes to have extra output parameters that can only be produced in grib2

PR #23, opened 2024-07-15
Detect ocean precision for PARKIND1_SINGLE_NEMO correctly

cxzk:
This improves the detection and configuration of coupling to a single-precision NEMO build.

We already do this correctly via the OCEAN_PREC variable for choosing the nemogcmcoup library to _link_ to, but the choice to define the preprocessor symbol PARKIND1_SINGLE_NEMO doesn't match this, and relies on the variable ${PNAME}_OCEANMODEL_HAVE_SINGLE_PRECISION ) which is used nowhere else (and doesn't seem to work reliably in all situations if you try to set it via bundle.yml.

If this is not correctly set, you end up with floating-point errors or segfaults in the coupling when REAL(8) WAM arrays get passed to NEMO routines expecting REAL(4) arrays (there are no interface blocks for this, so the compiler just assumes the data is correct rather than casting, leading to completely bogus values being interpolated in the coupling.

The proposed change simply determines whether to set PARKIND1_SINGLE_NEMO by checking whether OCEAN_PREC is SP - so that if we're linking against a single-precision NEMO, it will be set, and if we're not then it won't.


FussyDuck:
[CLA assistant check](https://bol-claassistant.ecmwf.int:443/ecmwf-ifs/ecwam?pullRequest=23) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://bol-claassistant.ecmwf.int:443/ecmwf-ifs/ecwam?pullRequest=23) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://bol-claassistant.ecmwf.int:443/check/ecmwf-ifs/ecwam?pullRequest=23) it.</sub>

cresswellp (approved):

awnawab (approved):
Thanks a lot for cleaning this up @cxzk, looks good to me! The macos CI test failures I think are unrelated; the runners appear to be down.

cxzk:
> This should be going into the develop branch, unless a hotfix release is required for this

That's fine - it will be needed for IFS CY50R1, along with the other (both scientific and technical) changes that need porting here from contrib/ecwam in IFS CY49R2.

However, this isn't super-urgent as we already have a temporary branch for early CY50R1 testing.

PR #25, opened 2024-08-02
IFS 49R2 science sync

awnawab:
This PR brings over all the scientific developments for the wave model as part of 49R2.

One of the main developments is the ability to output bathymetry files in grib, and this is now the default for etopo grids. When this option is enabled, the `wam_subgrid_?` files are output by `create_bathymetry` rather than by `preproc`. The scripts have been modified to reflect this; if `create_bathymetry` is run then the subgrid files are also cached for reuse just like `wam_topo`. The grib output is therefore disabled for aqua grids because these don't require `create_bathymetry`. We can potentially clean this up in a later PR if there is a case for needing grib output in aqua grids too.

I am filing this as a draft only because I have not updated all of the validation norms, just those needed for the CI. Once this PR is approved, I'll update the remaining validation norms.

Once merged, could this please be tagged as 1.4.0?


awnawab:
@jrbidlot and @jkousal32 could you also please review this PR? Thanks!

share/ecwam/scripts/ecwam_run_preproc.sh
```diff
@@ -22,6 +22,10 @@ source ${SCRIPTS_DIR}/ecwam_runtime.sh
 source ${SCRIPTS_DIR}/ecwam_parse_commandline.sh
 source ${SCRIPTS_DIR}/ecwam_helper_functions.sh
 
+if [[ $(uname) == "Darwin" ]]; then
+  export DR_HOOK_ASSERT_MPI_INITIALIZED=0
+fi
+
```
wdeconinck, line 28: In principle this export could be applied to all platforms, and can go into the ecwam_runtime instead.

share/ecwam/scripts/ecwam_run_model.sh
```diff
@@ -33,7 +33,7 @@ find_preproc_files
 find_preset_files ${begofrn}
 
 lgribin=F
-lgribout=F
+lgribout=T
```
wdeconinck, line 36: Hi @jrbidlot does this variable in the namelist affect the restart file output only?
Does this mean that the restarts are not preserving all bits anymore so that restarting is no longer bit-identical?

wdeconinck (commented):
Generally good to me. Thanks! And please check my question below.

share/ecwam/scripts/ecwam_run_model.sh
jrbidlot: This is correct. Writing out the restart file in gtib will not allow to have a restart that is bit identical.
We do it all the time when running under prepIFS because the restart files are then fetch with mars (hence in grib).
For the standalone testing, I don't think we actually continue a run from restart produced by a previous run but we need to assume that if users were to try to do this, they would restart from pure binary restart because they would not want to handle grib restart like we do ec ECMWF because it implies coding wave spectra in grib which no one does but ECMWF. So keep lgribout=F

cxzk (changes requested):
This generally looks consistent with my preliminary squashed merge that I've been using for testing (and indeed most differences from that are either in parts I didn't need to update for my purposes, or due to my own mis-merging.)

However, there are a couple of things that should be fixed - one is a (harmless) duplicate block of code that's been introduced in `OUT_ONEGRDPT`; the other is an apparent mis-merge in `WGRIB_ENCODE` that results in some spectral outputs not having `direction` and `frequency` set correctly - in particular this affects the bathymetry files generated by the IFS `wconst` task, which leads to a crash when the model tries to read them in.

Once that is fixed, this does seem to be bit-identical to my previous testing when run in the context of the full IFS continuous integration suite.

src/ecwam/out_onegrdpt.F90
```diff
@@ -151,6 +151,20 @@ SUBROUTINE OUT_ONEGRDPT(IU06)
...
+        LPARAM = .FALSE.
+      ENDIF
+
+      IPMSS=ITOBOUT(IRMSS)
+      IF(IPMSS.GT.0) THEN
+        LPARAM = .TRUE.
+      ELSE
+        LPARAM = .FALSE.
+      ENDIF
+
```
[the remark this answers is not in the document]
awnawab: Well spotted, thanks!

src/ecwam/wgribencode.F90
```diff
@@ -433,20 +491,20 @@ SUBROUTINE WGRIBENCODE ( IU06, ITEST, &
...
-        ELSE
-          CALL IGRIB_SET_VALUE(IGRIB_HANDLE,'indicatorOfUnitOfTimeRange',1)
-        ENDIF
+        CALL IGRIB_SET_VALUE(IGRIB_HANDLE,'indicatorOfUnitOfTimeRange',1)
         CALL IGRIB_SET_VALUE(IGRIB_HANDLE,'stepUnits','h')
         CALL IGRIB_SET_VALUE(IGRIB_HANDLE,'endStep',ISTEP_HRS)
 
       ENDIF
 
       IF (IPARAM == 251) THEN
```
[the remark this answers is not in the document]
awnawab: Thanks a lot for testing and finding this mis-merge 🙏

awnawab:
Many thanks all for your reviews and feedback 🙏 I've applied all of the fixes you suggested. I'll now update the remaining validation norms.

jrbidlot (approved):
As far as I can see this now inline with what went in 49R2.

cxzk:
> Many thanks all for your reviews and feedback 🙏 I've applied all of the fixes you suggested. I'll now update the remaining validation norms.

Thanks - the updated branch now passes all my testing for IFS CY50R1.

cxzk (approved):

awnawab:
Thanks again for the feedback all. I've now updated all the validation hashes, and I've also included a bug fix contributed by @jrbidlot offline.

wdeconinck (approved):
Looks great! Thanks!

PR #26, opened 2024-08-14
Template derived-types with fypp + yaml

awnawab:
The derived-types used to store state fields in `yowdrvtype.F90` followed a very regular pattern. After discussion with @jrbidlot and @jkousal32, these were fully-templated. Among other things, this made a deeper integration with FIELD_API much easier, allowing the actual memory allocations to be owned by FIELD_API rather than simply wrapping existing allocations. This will enable the use of powerful FIELD_API functionality such as host-side memory pinning (for faster  CPU-GPU data transfers). It should be noted that the templating is done in two layers simply to work around _excessively_ long compile times on nvhpc (~10 mins).

Other than that, this PR also rolls back some Loki related workarounds in both the source and the CI.


awnawab:
@jrbidlot and @jkousal32 could you also please review this PR? It has an impact on scientific developers of ecWAM too, as the derived-type definitions are now modified via the yaml config file.

jrbidlot (commented):
why do we need 2 different types, one for fl1_type and one for xllws_type?

The have the same dimensions, essentially dealing with the the spectral dimension (directions and frequencies) and the horizontal dimensions (block anf threads)

awnawab:
> why do we need 2 different types, one for fl1_type and one for xllws_type?
> 
> The have the same dimensions, essentially dealing with the the spectral dimension (directions and frequencies) and the horizontal dimensions (block anf threads)

Great point @jrbidlot! Both of these should belong in one instance of `4D_TYPE`.

jkousal32 (approved):
all looks good to me. Looking like the code is becoming increasingly easier and cleaner to work with - nice!

jrbidlot (approved):
As far as I can, the new derived types will become handy in future.
Well done

wdeconinck (approved):
Nice!  Great work!

PR #27, opened 2024-09-03
TAUT_Z0 convergence fix

awnawab:
As of 49R1, `LLGCBZ0` is set to `.TRUE.` and the point-wise convergence loop in `TAUT_Z0` is used. This was found to have a numerical instability that was triggered by the normal range of round-off error one can expect when running on GPU as compared to CPU.

@jrbidlot fixed the numerical instability and confirmed the change was meteorologically neutral. Many many thanks again for resolving this so quickly 🙏 🙏 

I am intentionally filing this PR to "develop-1.3", which is the technical development branch. "develop" is reserved for scientific developments, and this fix will make its way there through via a CY50R1 scientific contribution.

The PR also disables caching for the macos runners in the CI, this can sometimes be intermittently fragile.


awnawab:
Hi @jrbidlot. Could you please give this PR a quick look? I have just cherry-picked your commit and updated the validation hashes.

jrbidlot (approved):
Sorry Ahmad, I forgot to act.

It looks good.

PR #28, opened 2024-09-09
Fix to multio linkage from IFS CY49R1

cxzk:
This cross-ports a fix from IFS CY49R1 that was missing and its absence here was causing MULTIO configuration to fail in the standalone WAM experiment in the IFS develop CI.

* adapt CMakeLists.txt to latest multio-fapi


cresswellp (approved):
Strictly speaking, I'm not sure if both `HAVE_MULTIO` and `multio_FOUND` are necessary - this is internal ecBuild logic, but I suspect the former implies the latter.

@wdeconinck please can you advise - is the code here 'correct' or can it be simplified?

src/ecwam/CMakeLists.txt
```diff
@@ -397,15 +397,18 @@ endif()
 
 set(MULTIO_LIBRARIES)
 if( HAVE_MULTIO )
-  if( multio_HAVE_FDB5 )
-    # Force the linkage of multio-fdb5 for static builds
-    if( NOT BUILD_SHARED_LIBS )
-      list(APPEND MULTIO_LIBRARIES -Wl,--push-state,--no-as-needed multio-fdb5 -Wl,--pop-state)
-    else()
-     list(APPEND MULTIO_LIBRARIES multio-fdb5)
+  if(multio_FOUND)
```
awnawab, line 400: This is indeed superfluous, as we are already checking for `HAVE_MULTIO`.

src/ecwam/CMakeLists.txt
```diff
@@ -397,15 +397,18 @@ endif()
...
+    if(multio_HAVE_FDB5)
+      # Force the linkage of multio-fdb5 for static builds
+      if(NOT BUILD_SHARED_LIBS)
+        list(APPEND MULTIO_LIBRARIES -Wl,--push-state,--no-as-needed multio-fapi -Wl,--pop-state)
+      else()
+        list(APPEND MULTIO_LIBRARIES multio-fapi)
```
awnawab, line 406: Since the name of the library has changed, this would be a non backwards-compatible change that corresponds to a new multio release. Could we please add a multio version check as part of the`MULTIO` option?

awnawab (changes requested):
Many thanks for porting this over @cxzk! It looks sensible to me, I've left two small comments, the more important one being the version check as this will trigger a more easily debuggable configure time failure if we are using the wrong multio.

I'd also like to highlight that `develop-1.3` is a branch mainly targeted at GPU related developments on top of 49R1, one of which is not bitid.

Do you need this fix in a tag that is bitid to 49R1? If so, this should go in before #27 and should be tagged before that is merged.

cxzk:
> Do you need this fix in a tag that is bitid to 49R1? If so, this should go in before #27 and should be tagged before that is merged.

Yes, that is required in order to be able to get the standalone WAM CI test running as part of the IFS "develop" CI, as "develop" is only intended to take bit-identical changes relative to the current cycle - that's why I targeted it at develop-1.3 rather than develop.

src/ecwam/CMakeLists.txt
cxzk: No problem, I'll remove this extra check that was added in IFS CY49R1, if it's not actually required.

src/ecwam/CMakeLists.txt
cxzk: I'm not clear exactly what is required here to do that, but if you can advise then I'm happy to add this.

src/ecwam/CMakeLists.txt
wdeconinck: Agreed, multio_FOUND is not needed

src/ecwam/CMakeLists.txt
wdeconinck: Does this change depend on the version of multio?

src/ecwam/CMakeLists.txt
```diff
@@ -397,15 +397,18 @@ endif()
...
-    # Force the linkage of multio-fdb5 for static builds
-    if( NOT BUILD_SHARED_LIBS )
-      list(APPEND MULTIO_LIBRARIES -Wl,--push-state,--no-as-needed multio-fdb5 -Wl,--pop-state)
-    else()
-     list(APPEND MULTIO_LIBRARIES multio-fdb5)
+  if(multio_FOUND)
+    if(multio_HAVE_FDB5)
+      # Force the linkage of multio-fdb5 for static builds
+      if(NOT BUILD_SHARED_LIBS)
+        list(APPEND MULTIO_LIBRARIES -Wl,--push-state,--no-as-needed multio-fapi -Wl,--pop-state)
```
dsarmany, line 404: I don't think `--no-as-needed` is necessary when linking to `multio-fapi`.

src/ecwam/CMakeLists.txt
wdeconinck: This is tangential to this PR, but the static linking version
```bash
-Wl,--push-state,--no-as-needed multio-fdb5 -Wl,--pop-state
```
 is dated from 2017, and I think in the mean time we don't need this anymore because multio-fdb5 (at least from version 2.1 is always compiled as shared library.

Perhaps it is sufficient to just keep
```cmake
list(APPEND MULTIO_LIBRARIES multio-fapi)
```
So that means we must require multio VERSION 2.1

awnawab: It would be good to wrap up all the multio related cmake changes here if possible.

src/ecwam/CMakeLists.txt
cxzk: Removed in latest push.

src/ecwam/CMakeLists.txt
cxzk: Special static linking rule removed, and required version updated, as suggested, in latest push.

cxzk:
All comments hopefully addressed in latest push, which I can confirm is still working in the IFS context I was trying to fix.

awnawab (approved):
Many thanks @cxzk for addressing the comments, good to go now as far as I can tell 👍

wdeconinck (approved):
Thanks @cxzk for addressing all suggestions!

PR #29, opened 2024-09-12
Run all of `WAMINTGR` on GPU

awnawab:
This is the first of two PRs targeted at reducing the frequency of data transfers between host and device. As a first step, everything under `WAMINTGR` is ported to run on device. This will actually increase the amount of data transfers by moving them out of conditional clauses. The next PR will complete the work by hoisting the data transfers out of the advection loop in `WAMODEL`, leading to significant reductions in overall walltime.

In order to cut down on the FIELD_API related boilerplate in the model code, the derived-types are offloaded to device and they now own the field pointers that were previously local in `WAMINTGR_LOKI_GPU`.

The PR also includes a minor cleanup of the data offload directives in the wave propagation kernel.


wdeconinck (approved):
Looks good to me

mlange05 (approved):
Sorry for the delay. Yes, looks good and in line with the general Loki+FIELD API approach.

PR #31, opened 2024-09-24
CMake: Only search for OpenACC if the corresponding option is active

reuterbal:


awnawab (approved):
Looks good to me, thanks for adding this extra safety check! @wdeconinck could this also please be patched to 1.3.3 once merged? Apologies for the repeated patch requests 😅

CMakeLists.txt
```diff
@@ -127,16 +128,17 @@ if( ${CMAKE_VERSION} VERSION_LESS "3.25" AND (NOT DEFINED ENABLE_ACC OR ENABLE_A
     if( OpenACC_Fortran_FOUND AND OpenACC_C_FOUND )
         set( OpenACC_FOUND ON )
     endif()
```
wdeconinck, line 130: Not a problem of this PR, but good moment to tackle. The recent Cray compilers (at least on LUMI) only support OpenACC for Fortran, so that OpenACC_FOUND will be FALSE.
Perhaps we don't need the OpenACC::OpenACC_C target altogether.

reuterbal: Done

wdeconinck (approved):

PR #32, opened 2024-09-30
Update CI

awnawab:
This PR re-enables the nvhpc CI runner, adds SP tests and cleans up some of the Loki python version workarounds.


wdeconinck (approved):
Great, especially reenabling nvhpc!

PR #35, opened 2024-10-08
Enable GPU code paths if appropriate Loki transformation is used

awnawab:
A small bug fix to fix failing Loki CI tests.


reuterbal (approved):
Thanks, looks good!

PR #36, opened 2024-10-11
Update nvhpc version in CI

awnawab:
The CI nvhpc runners can be fragile. This PR updates the nvhpc version to 23.5, which will hopefully be more robust. The CI push trigger is also removed, as users can manually trigger the CI on specific branches should they need to outside of a PR.


wdeconinck (approved):
Why not trigger on pushes?

wdeconinck:
I think it's useful to test each push. so please yes :)

PR #37, opened 2024-10-23
Incorporate more WAM changes from IFS CY49R1

cxzk:
This incorporates more changes that were done in ecWAM in IFS CY49R1 (where it's still integrated into the IFS source) into the standalone version of ecWAM.

**NOTE: this is ideally needed in a develop-1.3.2 branch, to produce a 1.3.2.1 release, given that develop-1.3 and 1.3.3 are unfortunately no longer bit-identical to 1.3.2 which is what's used in IFS develop, in order to be able to do a clean IFS CY49R1 -> develop merge.**

Contains [IFS-3715](https://jira.ecmwf.int/browse/IFS-3715): bugfix erroneous output in oops


awnawab:
Hi @cxzk. Many thanks for porting over the fix 🙏  The GPU dev branch needed to stay on 49R1 science, but needed a non bitid but scientifically neutral fix to cure a numerical instability exposed when running on GPUs. I thought we wouldn't need any more fixes to CY49R1 and we might get away with having just two develop branches ("develop" is updated to 49R2 science). That's how we ended up without a develop branch that's bitid to 49R1.  I've now created develop-1.3.2. Could you please rebase your PR on that?

Secondly, rather than tagging a release for every fix, I would suggest ifs-bundle develop should track develop-1.3.2. And right before we need to release an ifs-bundle tag, we can tag ecwam too.

cxzk:
Thanks @awnawab - that's now rebased on develop-1.3.2

awnawab (approved):
Many thanks for porting over this fix! 🙏 The macos CI failures are unrelated to this PR and have something to do with pip. I'll investigate these separately.

cresswellp (approved):

wdeconinck (approved):

PR #40, opened 2024-11-08
Physics kernel GPU performance optimisation

awnawab:
This is the last of the GPU optimisation PRs and delivers a substantial boost to the performance of the source term computation on GPU. Many thanks again to @lukasm91 for showing us these optimisations!

The optimisations are mostly implemented as loki directives, but there are also a few changes to the source. The results are confirmed bit-identical in coupled IFS runs.

Once this is reviewed and merged, could this please be tagged as 1.3.4?


awnawab:
@jrbidlot and @jkousal32 could you also please review this PR?

src/ecwam/sinput_ard.F90
```diff
@@ -292,21 +295,25 @@ SUBROUTINE SINPUT_ARD (NGST, LLSNEG, KIJS, KIJL, FL1, &
          CALL ABORT1
       ENDIF
 
-      DO IGST=1,NGST
-        DO IJ=KIJS,KIJL
-          USTPM1(IJ,IGST) = 1.0_JWRB/MAX(USTP(IJ,IGST),EPSUS)
-        ENDDO
+      DO IGST=1,2
+        IF(IGST <= NGST)THEN
```
jkousal32, line 299: this doesn't look very intuitive to me, just in terms of readability. Does NGST do anything then? i.e. if I were to change it to NGST=3?

awnawab: Yes that's a good point, as increasing NGST would now also require you to increase the loop bounds, and for a simple change you would have to change too many lines. Ok, let me double check but I think this particular change only has a small performance impact and we can revert it.

lukasm91: Another option would be to implement something that can generate specific kernels for NGST=1 and NGST=2, but that ends up with combinatorial explosion.

jkousal32: maybe I'm missing something, but why do we need the following line?
        IF(IGST <= NGST)THEN
is this not always true then?

lukasm91: depending on the chosen config, NGST is 1, or 2. We just convince the compiler that there is an upper bound. I agree that this is not very obvious, so either a comment or also a "loki-pass" (not sure how that would look like) might be worth to consider.

jkousal32: ok, thanks both for the information.

awnawab: Loki will already demote temporaries like `TAUX` because of the constant in the declaration. That combined with all of the other optimisations means that the hit from not unrolling `NGST` is just ~1%. Possibly once we've developed loki functionality to apply some of your (@lukasm91) more invasive changes (e.g. splitting the source-term kernel), this might become more important again. Until then, I'll revert this change. Thanks both for the discussion 😄

jrbidlot (changes requested):
I have left a few comments
I know it is a bit of extra work but adapting sinput_jan in the same way as sinput_ard will make sure that the option IPHYS=0 is kept working well

src/ecwam/sinput_ard.F90
awnawab: Ah, if we don't intend on increasing NGST above 2, then it makes sense to retain Lukas' loop unrolling optimisation. And yes of course I can add clarifying comments. Thanks for the clarification @jrbidlot!

src/ecwam/sinput_jan.F90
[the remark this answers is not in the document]
awnawab: Yes good spot, I will also apply the same optimisations to `SINPUT_JAN`. The CI already runs a unit-test with `IPHYS=0`, so we can be sure the optimisations are safe. I won't however test the `SINPUT_JAN` change in coupled runs unless you tell me that `IPHYS=0` is also an important configuration for the IFS.

src/ecwam/snonlin.F90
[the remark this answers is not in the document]
awnawab: The optimisations in snonlin indeed had a significant impact. In fact, @lukasm91 also demonstrated a clear benefit of placing snonlin in a separate driver loop so that it can be compiled as a separate GPU kernel allowing the compiler to better allocate registers. We are working on Loki functionality to automate the splitting of routines like `IMPLSCH`.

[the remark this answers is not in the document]
jrbidlot: Good to know. The current version of snonlin is an approximation that has been with 3rd generation wave models for over 30 years. More accurate alternatives exists and knowing that we might be able to use GPU kernel for them might allow research in that direction.

src/ecwam/sinput_jan.F90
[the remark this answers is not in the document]
jrbidlot: The CI test will be good enough.

awnawab:
Thanks all for the feedback. As requested, I've added an explanatory comment on the `SINPUT_ARD` loop unrolling optimisation, and have also extended this to `SINPUT_JAN`.

jrbidlot:
It looks good to me.

jrbidlot (approved):

jkousal32 (approved):

mlange05 (approved):
Looks very(!) good to me. Many thanks, GTG. 👍

wdeconinck (approved):

PR #41, opened 2024-11-21
Cache subgrid files in preproc

awnawab:
The output of the model is very sensitive to the bathymetry and obstruction coefficients. Previously, these were both computed by create_bathymetry, and the resulting bathymetry files were cached in nexus (https://get.ecmwf.int/#browse/browse:ecwam:data%2Fbathymetry%2Fv1%2Fbathymetry_O1280_nfre29_ETOPO1_e6c62ffa77b1815082065ffb241be879.gz). Preproc would then read this bathymetry file and generate the subgrid files. Repeatable results across different compilers and architectures were thus achievable as long as the same bathymetry file was used.

Now that grib output is enabled, create_bathymetry rather than preproc computes the grids, and outputs the subgrid files in grib. Therefore the subgrid files now also have to be cached in nexus in order to get repeatable results across different compilers. This PR makes the necessary changes to the scripts to achieve this.

Right now, I have only uploaded the bathymetry and subgrid files for the O48 grids to nexus. As this is a new caching approach they are in the v2 folder. Older versions of ecwam that don't use grib output will continue to use the bathymetries in the v1 folder. Once the PR is approved I will upload the bathymetries and subgrids for the O320, O640 and O1280 grids too.

This PR also adds single precision validation hashes for the O320 test case.

### DIFF
--- a/share/ecwam/scripts/ecwam_run_create_bathymetry.sh
+++ b/share/ecwam/scripts/ecwam_run_create_bathymetry.sh
@@ -136,11 +136,13 @@ fi
 
 cat reference_levels >> for_md5
 md5=$(cat for_md5 | md5sum | awk '{print $1}')
-WAM_TOPO=v${ecwam_bathymetry_version}/bathymetry_${wamresol}_nfre${wamnfre}_${wambathy}_${md5}
+
+# Only double-precision files are cached
+WAM_TOPO=v${ecwam_bathymetry_version}/bathymetry_dp_${wamresol}_nfre${wamnfre}_${wambathy}_${md5}
 
 subgrid_files=()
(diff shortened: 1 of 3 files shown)


awnawab:
@jrbidlot and @jkousal32 could you also please take a look and tell me if you agree with this change to the runner scripts?

jrbidlot (approved):
Hi Ahmad, good point that now with grib format of the files, we can cached them and make sure only one flavour is ever used (dp).
The intention of moving to grib format was always to be more portable.

jkousal32 (approved):

wdeconinck (approved):
Thanks @awnawab after our discussions offline everything is clear and good to go

PR #43, opened 2024-11-26
Merge develop-1.3 into develop

awnawab:
This PR merges the GPU offload developments from develop-1.3 into develop. Other than the numerical stability fixes to the gravity-capillary model in #27, the results are bit-identical to 1.4.0, which was confirmed via the following testing:
- Coupled IFS SP/DP runs with Intel, coupled IFS DP run with nvhpc
- Bit-identical statistics output for all 87 output parameters enabled
- Offline testing for the finer test configs: O320, O640, O1280


wdeconinck:
Very nice culmination of a lot of work finally making it into develop! Thanks @awnawab ! 🙏

wdeconinck (approved):

PR #45, opened 2024-11-28
Enable GPU offload of depth and current refraction solver

awnawab:
49R2 enables the current and depth refraction solver in the wave propagation kernel. This PR ports this solver to run on GPU. The CPU and GPU variants posed conflicting demands for performance, so I am now also using Loki to preprocess `PROPAGS2` for GPU enabled runs.


awnawab:
@jrbidlot and @jkousal32, could you also please have a look? Especially the new test config which enables current refraction, I am not sure I have configured this correctly.

tests/etopo1_oper_an_fc_O320_cy49r2.yml
jrbidlot, line 1: Can I ask how you chose the value of nproma?

awnawab: I set it based on GPU performance, since these test cases have mostly been used to assess GPU performance. If you know of a better value for CPU performance then I am happy to change it to that.

share/ecwam/scripts/ecwam_run_model.sh
jrbidlot, line 1: Until we actually supply the currents which will contain actual surface current data, it does not matter what we specify for CDATECURA because the routine that reads the potential file called currents, on not finding it will simply initialise the currents to 0.
But in preparation on when we will actually add the file currents, I would suggest to set 
CDATECURA             = "${begofrn}",
i.e. we will expect current data from the beginning of the run.
We should also add the namelist entry IDELCUR. It will tell ecWAM how often in seconds the file currents will contain new data. For now we can select IDELCUR = 86400,
(i.e. new current field every day)

jrbidlot (changes requested):
See my comment on adding IDELCUR and future work on adding actual ocean current data

tests/etopo1_oper_an_fc_O320_cy49r2.yml
jrbidlot: We have used nproma values around 24 but actually, it has not been really tested on Atos to determine if there is another optimum one.

jkousal32 (approved):

awnawab:
Thanks a lot @jrbidlot for the feedback! I've addressed your comments, please have another look when you can 😄

jrbidlot (approved):
Looks fine to me.
Thanks

mlange05 (approved):
GPU and Loki changes look good to me. 👍

wdeconinck (approved):

PR #47, opened 2024-12-10
Add default compiler flags for Cray

ioanhadade:
Add compiler flags -ram -emf -hflex_mp=conservative -Othread1 -hfp1 -hadd_paren to prevent random SIGFPEs


cmake/ecwam_compile_flags.cmake
```diff
@@ -16,6 +16,7 @@ if(CMAKE_Fortran_COMPILER_ID MATCHES "Cray")
   set(checkbounds_flags   "-Rb")
   set(fpe_flags           "-Ktrap=fp")
   set(initsnan_flags      "-ei")
+  set(general_flags       "-ram -emf -hflex_mp=conservative -Othread1 -hfp1 -hadd_paren" )
```
awnawab, line 19: At the time I was suggesting a quick fix, but I think we can split up these flags into more descriptive categories. So possibly something like this:

```
set(threading_flags "-hflex_mp=conservative -Othread1")
set(fpmodel_flags "-hfp1 -hadd_paren")
set(baseline_flags "-ram -emf")
```

awnawab (changes requested):
Thanks a lot @ioanhadade for filing this fix 🙏  just a small nitpicky comment that would help clarify what some of these flags are for.

awnawab (approved):
Thanks a lot @ioanhadade for addressing the comment 🙏 one optional nitpick is to remove the setting of `general_flags` which no longer exist, but I can also mop this up in a future PR.

ioanhadade:
> Thanks a lot @ioanhadade for addressing the comment 🙏 one optional nitpick is to remove the setting of `general_flags` which no longer exist, but I can also mop this up in a future PR.

I can do it now. One sec.

PR #51, opened 2025-01-15
Loki: Remove deprecated `DIRECTIVE` option in Loki CMake macro call

mlange05:
This is about to be removed upstream as part of the entry-point clean-up.


awnawab (approved):
Thanks a lot for filing this fix, looks great 👌

PR #52, opened 2025-01-15
Add fp-model precise to stresso.F90 to prevent SIGFPEs with newer Intel compilers

ioanhadade:
Newer Intel compilers (e.g., 2021.10.0) raise a SIGFPE in stresso.F90. A quick fix is to build that file with fp-model=precise which seems to be performance neutral. This works also with older Intel compilers and with newer ones.


awnawab (approved):
Thanks a lot for filing this fix, looks great 👌

PR #60, opened 2025-01-21
Open input files with read-only rather than readwrite to prevent lock storms

ioanhadade:
When run in coupled mode, we open wam_grid_tables and other binary files as readwrite even though we only read from them. This can cause lock storms as in some places we do this operation with all tasks and also prevents sharing of inputs since we need to set rw permissions for groups or others. If run in standalone mode, where some of these files are written to, the open semantics does not change.


awnawab (approved):
Thanks a lot for this fix, looks good to go 👌

PR #62, opened 2025-02-04
Port across fix from IFS CY49R1 (IFS-3914)

cxzk:
From [IFS-3914](https://jira.ecmwf.int/projects/IFS/issues/IFS-3914):
* fix for infinite loop in getspec when reading input data on a different resolution than the model one


awnawab (approved):
Thanks for bringing over this fix @cxzk! Looks good to me 👌

PR #64, opened 2025-02-06
Add hpc-ci with CPU and temporarily disable hpc-ci with GPU until GPU offload is reenabled

awnawab:


awnawab:
As agreed offline, I have added a cpu hpc-ci build and only temporarily disabled the gpu build.

wdeconinck:
That's great, at least it will be running on the hpc with CPU  :)

PR #65, opened 2025-02-06
Remove hard-coded static linking for GPU builds

awnawab:


wdeconinck:
Would we not like all libraries including dependencies to be statically linked in case of --with-static-linking ?

awnawab:
As agreed offline, I have extended the static linking to all dependencies and also added a cpu hpc-ci build.

PR #72, opened 2025-03-07
1.5.0 wave sea ice

jkousal32:
the passive and bit-identical implementation of all the code required for more advanced wave-sea ice coupling (see here for outline: https://jira.ecmwf.int/browse/IFS-3634). 
-     implement effect of waves on sea ice (waves break up the sea ice, and wave stress to sea ice)
-     will include new prognostic field ice breakup memory in the sea ice model
-     wave stress to sea ice will also require two new fields for the x,y components

This must be pulled in consecutively with
- ifs-source: pajk_CY49R1_for_50R1_v12_waveSeaIce.IFS-3634
- ifs-scripts: pajk_CY49R1_for_50R1_v12_waveSeaIce.IFS-3634
- ifs-defaults: pajk_CY50R1_v12_waveSeaIce.IFS-3634

checks for bit-identicality have been done, and it is likely that this will be able to go into 50r1 passively


jkousal32:
@jrbidlot can you also review this? these are the associated wave changes for my recent IFS PR

src/ecwam/sdice2.F90
```diff
@@ -75,6 +75,7 @@ SUBROUTINE SDICE2 (KIJS, KIJL, FL1, FLD, SL,            &
 
       REAL(KIND=JWRB), DIMENSION(KIJL,NANG,NFRE), INTENT(IN) :: FL1
       REAL(KIND=JWRB), DIMENSION(KIJL,NANG,NFRE), INTENT(INOUT) :: FLD, SL
+      REAL(KIND=JWRB), DIMENSION(KIJS,NANG,NFRE), INTENT(OUT) ::        SLICE
```
awnawab, line 78: Shouldn't this be `KIJL`?

src/ecwam/sdice3.F90
```diff
@@ -75,6 +76,7 @@ SUBROUTINE SDICE3 (KIJS, KIJL, FL1, FLD, SL,            &
 
       REAL(KIND=JWRB), DIMENSION(KIJL,NANG,NFRE), INTENT(IN) :: FL1
       REAL(KIND=JWRB), DIMENSION(KIJL,NANG,NFRE), INTENT(INOUT) :: FLD, SL
+      REAL(KIND=JWRB), DIMENSION(KIJS,NANG,NFRE), INTENT(OUT) ::        SLICE
```
awnawab, line 79: Same comment as for sdice2.

awnawab (changes requested):
I can't really comment on the science changes, but just had a question about two of the dummy argument size specifications. Other than that, looks good to me 👌

src/ecwam/sdice2.F90
jkousal32: yes indeed, thanks for that

src/ecwam/wdfluxes.F90
jrbidlot, line 1: same comment as in implsch

src/ecwam/sdice3.F90
jrbidlot, line 1: Now that the level of impliciteness factorXIMP is used in more than one places, it should be defined module (yowstat) so that if it were to be changed it would be everywhere.

src/ecwam/implsch.F90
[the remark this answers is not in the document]
jkousal32: hmm. the reason I did this was actually that for analysis type experiment with **LWNEMOCOUWRS=true**, it failed in 
OUTSTEP0 -> WDFLUXES -> WNFLUXES
on the following line
SUMXICE(IJ) = SINTH(K)*MIN(SLICE(IJ,K,M),-EPSMIN1000)
my reasoning was that SLICE was not defined here because SDICE had not yet been called at step0, which actually doesn't seem to make sense to what I coded into the logic... (assuming all switches are set properly, SDICE should always be called before WNFLUXES)
this issue was fixed by initializing it as I've done here. 
I guess something fishy is going on here, and I will have a look into it when I am back for mission next week

(this is testing of the further coupled switches, nothing relevant to the current 50r1 cycle)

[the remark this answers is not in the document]
awnawab: In `WNFLUXES`, `SLICE` is `INTENT(IN)` and so in the case we don't call `SDICE`, we should zero it in `IMPLSCH` and `WDFLUXES`. I think Jean's suggestion is a good one, and we can save an unnecessary write by only zeroing `SLICE` if `SDICE` is not called.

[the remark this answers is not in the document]
jkousal32: this is now fixed

src/ecwam/sdice3.F90
jkousal32: yes, actually, on that:

this is some part of your code which I've now copied around a few times

DELT = IDELT
DELTM = 1.0_JWRB/DELT
XIMP = 1.0_JWRB
DELT5 = XIMP*DELT
...
which is then later used in the following way
...
!              to be used for wave radiative stress calculation
               GTEMP1         =  MAX((1.0_JWRB-DELT5*FLDICE),1.0_JWRB)    
               SLICE(IJ,K,M)  =  SLICE(IJ,K,M)/GTEMP1

 Can I thin it down to the following, removing XIMP? or do we need this to be possibly changed at some point, and want to keep it as a variable?
 
DELTM = 1.0_JWRB/IDELT
DELT5 = IDELT

I note that these changes will again mean non bit-identicality, so would probably not introduce them now, but in my development branch.

jrbidlot: NO. I know it seems silly to multiply by 1 but XIMP is an important variable as it represent the level of implicitness in the time integration for the physics. The original WAM had 0.5. I cannot rule out that one day we might want to use a different value than 1.

PR #73, opened 2025-03-07
More CY49R1 fixes

cxzk:
This ports across the latest CY49R1 fixes from ifs-source, corresponding to https://git.ecmwf.int/projects/IFS/repos/ifs-source/pull-requests/1440

Needed on the develop-1.3.2 branch for IFS "develop" which should remain bit-identical to that release.


awnawab (approved):
Thanks @cxzk for porting over these fixes. As discussed offline, I'll make sure this makes its way to the 49R2 and 50R1 branches too.

PR #74, opened 2025-03-14
Limit updates of surface currents when provided by NEMO

jrbidlot:
This set of changes are only relevant when ecWAM is coupled to NEMO. In that case, surface currents will be provided every hour. However, without the changes, ecWAM will process the currents every time it is coupled to IFS which is less than 1 hour. With the changes, the processing will be limited to when new surface currents are provided.
This is a bit-identical change.


jrbidlot:
@jkousal32 and @awnawab not sure if it is the way to trigger a review. In any case, can you review the changes as they will improve performance for CY50R1

awnawab (approved):
Thanks @jrbidlot, this looks great! We're in the process of rolling out github enterprise IDs, but until then yes sadly the only way to trigger reviews is like you have done. As for the one failing CI test, that's an artefact of a fix in the CI setup for runs on our hpc that was contributed after 1.5.0 was tagged. So please ignore that for now, and any PRs on a future 1.5.1 tag won't have this problem.

jkousal32 (approved):
looks good!

wdeconinck (approved):

PR #75, opened 2025-03-17
Fix finding of python modules

awnawab:
A small PR that fixes the finding of python modules if an alternative python interpreter is specified via `-DPython3_ROOT_DIR`.

I've also piggybacked a bitid fix to 49R1 contributed in #73.


src/ecwam/userin.F90
```diff
@@ -1604,7 +1604,7 @@ SUBROUTINE USERIN (IFORCA, LWCUR)
             LERROR = .TRUE.
           ENDIF
         ENDDO
-      ELSE
+      ELSE IF (.NOT.LRESTARTED) THEN
```
wdeconinck, line 1607: A final `ELSE` is missing perhaps?
```Fortran
IF ( ... ) THEN
...
ELSE IF ( ... ) THEN
...
ELSE
! this is not covered !
ENDIF
```

awnawab: Well spotted! But this was merged as is in CY49R1_develop, so we should also merge this as is here. If needed we can then file a bugfix PR to develop.

wdeconinck (approved):

PR #79, opened 2025-04-10
Fix name of develop branch in pr-check workflow (was staging)

msleigh:
Sorry @awnawab - I didn't spot that I was using `staging` as the branch name in the workflow file, whereas the branch name in this repo is `develop`.

### DIFF
--- a/.github/workflows/pr-check.yml
+++ b/.github/workflows/pr-check.yml
@@ -1,7 +1,7 @@
-name: Check Pull Request
+name: check-pr
 
 on:
-  pull_request:
+  pull_request_target:
     types: [opened, synchronize, reopened]
     branches: [ main, develop ]
 
@@ -16,6 +16,8 @@ jobs:
 
       - name: Checkout code
         uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.base.sha }}
 
       - name: Set up uv
         uses: astral-sh/setup-uv@v5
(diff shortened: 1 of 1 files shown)


awnawab:
Great catch! The workflow now seems to be failing with this error: `fatal: could not read Username for 'https://github.com/': terminal prompts disabled`. Any idea what could be causing this?

msleigh:
> Great catch! The workflow now seems to be failing with this error: `fatal: could not read Username for 'https://github.com/': terminal prompts disabled`. Any idea what could be causing this?

No idea, this is mystifying me. Basically the PRTK_READ_ACCESS_TOKEN environment variable is unset. The same workflow file is used in ifs-git-tools and is identical, to the letter (apart from the staging/develop branch name difference), and it works fine. The PRTK_READ_ACCESS secret is defined at the Organization level, and is available to all repositories in the Org, so all should work indentically. I'm currently at a loss to see why this repo can't read secrets.PRTK_READ_ACCESS.

awnawab:
Could it be an artefact of staging the branch on your fork rather than directly on the central repo? I have seen funny effects in the past related to staging github CI related branches on forks, and I now tend to stage CI related branches on the central repo.

msleigh:
> Could it be an artefact of staging the branch on your fork rather than directly on the central repo? I have seen funny effects in the past related to staging github CI related branches on forks, and I now tend to stage CI related branches on the central repo.

Yes, could be, although I staged the branch from my fork when I did this with ifs-git-tools too and that worked. I could also now I think about it be due to the fact that ecwam is public, whereas ifs-git-tools is private, so maybe there are further restrictions on use of secrets.

PR #80, opened 2025-04-10
Fix name of develop branch in pr-check workflow (was staging)

msleigh:
Trying the same PR as #79 but within ecwam repo rather than from fork


msleigh:
Redundant; duplicate of #79 for testing only

PR #90, opened 2025-05-14
Pin ecbuild to 3.10.0 to fix failing CI

awnawab:
One of the recent merges to ecbuild broke ifort builds. This PR pins the version of ecbuild used in the CI to the last point release whilst the ecbuild bug is investigated further.

### DIFF
--- a/.github/workflows/build-hpc.yml
+++ b/.github/workflows/build-hpc.yml
@@ -106,7 +106,7 @@ jobs:
                   - -DENABLE_FCKIT_VENV=ON
                   - -DBUILD_SHARED_LIBS=${{ !matrix.gpu }}
               ecmwf-ifs/fiat:
-                version: 1.4.1
+                version: 1.5.1
                 cmake_options:
                   - -DENABLE_MPI=ON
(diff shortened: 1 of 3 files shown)


wdeconinck:
The issue is actually present in fiat, and was exposed by a fix in ecbuild.
The fix in fiat has been pushed to develop. I will create a hotfix fiat release for fiat, and the alternative could be to then use the newest fiat release.

awnawab:
Ah right, thanks for the clarification! Yes fixing the root cause would be much better. Could you please roll a fiat hotfix tag then and I'll update this PR accordingly?

awnawab:
I've rolled back the ecbuild change, and instead updated the FIAT version, as you suggested. The failing hpc-ci tests are due to an update of the default python3 module in the atos, which is incompatible with loki v0.2.9. I've updated the hpc-ci workflow to use python3.11, but for PRs filed from forks, the hpc-ci uses the target branches workflow, so the python version change won't come into effect until this is merged.

I have tested it offline that using python3.11 does indeed work, so please merge this unless you see any other outstanding issues.

wdeconinck (approved):

PR #93, opened 2025-06-10
IFS-4177: Single bundle invocation for ecWAM

awnawab:
Building on top of the functionality contributed in #70, this PR adds the ability to build both precisions of ecWAM with a single ecbundle invocation. The ability to mix and match ecWAM precision with nemo precision is also retained. Backwards compatibility is maintained with older bundles which still have two sp/dp symlinked entries for ecWAM. However the source still needs a small change, as wam.cmake creates symlinks to the ecWAM executables which now have different names.


src/ecwam/CMakeLists.txt
```diff
@@ -452,7 +459,7 @@ foreach( prec sp dp )
         PUBLIC_INCLUDES  $<INSTALL_INTERFACE:include>
         PRIVATE_INCLUDES ${CMAKE_CURRENT_SOURCE_DIR}
         PRIVATE_DEFINITIONS ${ECWAM_PRIVATE_DEFINITIONS} $<${HAVE_CUDA}:WAM_HAVE_CUDA>
-        PUBLIC_DEFINITIONS ${ECWAM_DEFINITIONS}
+        PUBLIC_DEFINITIONS ${ECWAM_DEFINITIONS} $<$<STREQUAL:${_OCEAN_PREC},SP>:PARKIND1_SINGLE_NEMO>
```
wdeconinck, line 462: Can PARKIND1_SINGLE_NEMO be added to the PRIVATE_DEFINITIONS ?

wdeconinck (commented):
Looks good. Just one question, see below.

PR #92, opened 2025-05-26
Naan update gpu offload

awnawab:

PR #96, opened 2025-06-17
Restore GPU offload

awnawab:
This PR restores GPU offload for the current tip of develop. The improvements contributed in #92 are also included herein.


mlange05 (approved):
Haven't tested this myself, but glad it re-enables the GPU-test. Otherwise looks good to me. 👍

PR #97, opened 2025-06-18
Adding mars stream clte  option

stefaniereuter:
This commit enables destine experiments to be setup with the correct stream settings and avoids doing this in postprocessing. As part of the grib migration IFS streams and wave streams will be identical. This was already applied in Destine, therefore the clte stream ISTREAM does map to the same KSTREAM. 

Up until know every case here sets also the LASTREAM variable, which is now misleading as they are both IFS stream and Wave stream.

### DIFF
--- a/src/ecwam/wstream_strg.F90
+++ b/src/ecwam/wstream_strg.F90
@@ -528,6 +528,12 @@
         MARSFCTYPE = 'fc'
         KSTREAM = 1248
         LASTREAM=.TRUE.
+      ELSE IF(ISTREAM.EQ.1098) THEN
+!       WAVE DESTINE FORECAST
+        CSTREAM = 'clte'
+        MARSFCTYPE = 'fc'
+        KSTREAM = 1098
+        LASTREAM=.TRUE.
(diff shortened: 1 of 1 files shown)


FussyDuck:
[CLA assistant check](https://bol-claassistant.ecmwf.int:443/ecmwf-ifs/ecwam?pullRequest=97) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://bol-claassistant.ecmwf.int:443/ecmwf-ifs/ecwam?pullRequest=97) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://bol-claassistant.ecmwf.int:443/check/ecmwf-ifs/ecwam?pullRequest=97) it.</sub>

jkousal32 (approved):
looks good. can confirm non-invasive. KSTREAM is used only for this singular purpose.

awnawab (approved):
Thanks a lot @stefaniereuter for filing this fix! And thanks also for bearing with me through the rebase and CI mix up, now G2G!

PR #99, opened 2025-07-14
Ecwam 1.5.0 phioc fix

jkousal32:
Two modifications:

1. a bug in the diagnostic calculation for step0 (wdfluxes.F90) (scaling for sea ice concentration was applied differently between the different source terms). This is important because this step0 field is also used in forcing the operational ocean system (OLDA stream)

3. A modification of how the masking is done and default values are applied when LWAMRSETCI=true (wnfluxes.F90)

These are both bit-identical changes because the changes in 1) are diagnostic currently, and the changes in 2) because LWAMRSETCI=false in 50r1

phioc before (left) and after normalization (right)
with bug:
Screenshot 2025-07-14 at 14 39 01
after bugfix:
Screenshot 2025-07-14 at 14 39 07


src/ecwam/wdfluxes.F90
```diff
@@ -93,30 +93,30 @@ SUBROUTINE WDFLUXES (KIJS, KIJL,                         &
...
+      REAL(KIND=JWRB), DIMENSION(KIJL), INTENT(INOUT) :: UFRIC, Z0M, Z0B, CHRNCK, CITHICK
+      REAL(KIND=JWRB), DIMENSION(KIJL), INTENT(INOUT) :: TAUXD, TAUYD, TAUOCXD, TAUOCYD, TAUOC
+      REAL(KIND=JWRB), DIMENSION(KIJL), INTENT(INOUT) :: TAUICX, TAUICY
+      REAL(KIND=JWRB), DIMENSION(KIJL), INTENT(INOUT) :: PHIOCD, PHIEPS, PHIAW, USTOKES, VSTOKES
+      REAL(KIND=JWRB), DIMENSION(KIJL), INTENT(INOUT) :: WSEMEAN, WSFMEAN, STRNMS
+      REAL(KIND=JWRO), DIMENSION(KIJL), INTENT(INOUT) :: NEMOUSTOKES, NEMOVSTOKES, NEMOSTRN
+      REAL(KIND=JWRO), DIMENSION(KIJL), INTENT(INOUT) :: NPHIEPS, NTAUOC, NSWH, NMWP, NEMOTAUX
+      REAL(KIND=JWRO), DIMENSION(KIJL), INTENT(INOUT) :: NEMOTAUY, NEMOWSWAVE, NEMOPHIF
+      REAL(KIND=JWRO), DIMENSION(KIJL), INTENT(INOUT) :: NEMOTAUICX, NEMOTAUICY
+      REAL(KIND=JWRB), DIMENSION(KIJL,NANG,NFRE), INTENT(OUT) :: XLLWS
```
cxzk, line 119: It's not clear to me how the DIMENSION changes here relate to the substance of this PR?

cxzk (approved):
I don't consider that I'm in a position to comment on the scientific correctness of the change, but if it's agreed that this is appropriate for 50r1 then I'm fine with the merge.

src/ecwam/wdfluxes.F90
jrbidlot: It is not directly related to the scientific merit of the PR but this change on dimension is inline with what had to to be done when the code was adapted to GPU and it shoudl have been adapted to that standard.

awnawab (changes requested):
Many thanks for this fix @jkousal32 and @jrbidlot. While the changes look fine, there are some conflicts with develop, which has seen quite a few bitid fixes since 1.5.0 was tagged. Could you please rebase the branch over develop? you can also merge develop into your branch if you prefer (rebasing does give a cleaner git history though 😉 )

jkousal32:
this has been moved to https://github.com/ecmwf-ifs/ecwam/pull/101 (rebased onto develop)

PR #100, opened 2025-07-22
Restore SCC vector transformations

awnawab:
A small PR to restore the Loki scc vector transformations, bringing the naming convention in ecwam in line with the IFS.


awnawab:
I've also piggybacked a small fix to avoid spurious FPEs in NVHPC for coarse resolution IFS coupled runs.

jkousal32 (approved):

mlange05 (approved):
Looks good to me. GTG 👍

PR #101, opened 2025-07-22
Phioc fix

jkousal32:
Two modifications:
(note, this is a duplicate PR of https://github.com/ecmwf-ifs/ecwam/pull/99, but rebased onto develop)

1. a bug in the diagnostic calculation for step0 (wdfluxes.F90) (scaling for sea ice concentration was applied differently between the different source terms). This is important because this step0 field is also used in forcing the operational ocean system (OLDA stream)

2. A modification of how the masking is done and default values are applied when LWAMRSETCI=true (wnfluxes.F90)

These are both bit-identical changes because the changes in 1) are diagnostic currently, and the changes in 2) because LWAMRSETCI=false in 50r1

phioc before (left) and after normalization (right)
with bug:
Screenshot 2025-07-14 at 14 39 01
after bugfix:
Screenshot 2025-07-14 at 14 39 07

PLUS! a new bug fix:
- regarding how the subgrid bathymetry information is output in grib2 for configuration that are in lat-lon over a subarea of the globe (nothing we use in operations or even in the basic tests in ecwam).


src/ecwam/wdfluxes.F90
```diff
@@ -213,6 +214,27 @@ SUBROUTINE WDFLUXES (KIJS, KIJL,                         &
...
+        CALL SNONLIN (KIJS, KIJL, FL1, FLD, SL, WAVNUM, DEPTH, AKMEAN)
+
+        IF (LWVFLX_SNL) THEN
+!         Save source term contributions relevant for the calculation of ocean fluxes
+!!!!!!    SL must only contain contributions contributed to fluxes into the oceans
+!         MODULATE SL BY IMPLICIT FACTOR
+          DO M=1,NFRE
+            DO K=1,NANG
+              DO IJ=KIJS,KIJL
+                GTEMP1 = MAX((1.0_JWRB-DELT5*FLD(IJ,K,M)),1.0_JWRB)
```
awnawab, line 238: GTEMP1 and DELT5 were deleted in this commit: https://github.com/ecmwf-ifs/ecwam/commit/53158275608eefbe35674b7346215088e4f6aff6. So these need to be restored.

awnawab (approved):
Many thanks @jkousal32 for the rebase, this is now G2G 👌 @cxzk is this supposed to be included in 50R1? If yes, I guess you need a new tag?

cxzk:
Yes please, if you could tag a new release containing this for use in 50r1.

PR #102, opened 2025-07-29
Sync with CY49R3_v4

ekuw:
Sync ecwam with CY49R3_v4.
DE-2805, DE-2842, IFS-4284
This update includes modifications for GRIB2-only running from CY49R2_v9 (which still has ecwam in the bundle). This means that these updates would be incompatible with CY49R2b, and area only appropriate for GRIB2-only running.
It also includes a few patches introduced to get the CI experiments running for CY49R2_v9.


src/ecwam/outwspec.F90
```diff
@@ -175,6 +175,8 @@ SUBROUTINE OUTWSPEC (IJS, IJL, SPEC, MARSTYPE, CDATE, CDATED, IFCST)
 ZRECVBUF(:)=0._JWRB
 
 ISTEP=NPROC
+!!!debile
```
awnawab, line 178: Hi @jrbidlot. Is this a debug leftover? If so, should/can this be removed?

awnawab (approved):
Thanks a lot @ekuw for filing this patch 🙏  Looks good to me 👌 I just had one small comment about what looks like a debug leftover, but this is probably harmless.

src/ecwam/outwspec.F90
jrbidlot: I am not sure whether line 179 is needed for what has been done with 49R3.
it is something I would have added but I don't remember if it is really needed

awnawab: Ok. In that case if the current branch has been tested successfully in 49R3 then we can leave it as it is.

jrbidlot:
Ahmad,
to be sure these changes will not go towards the release we use for 50R1 as some of the changes are for only grib2 running

jkousal32 (approved):
looks good to me!

awnawab:
> Ahmad, to be sure these changes will not go towards the release we use for 50R1 as some of the changes are for only grib2 running

Thanks for the clarification Jean, I'll remember to keep this patch out of develop.

PR #103, opened 2025-07-30
Small offload fixes

awnawab:
This PR contributes the following two offload related fixes:
- Add `finalize` clause to ensure device allocation for MPI buffers is always deleted
- Move the allocation of `WVPRPT_LAND` to `WVALLOC` so that it can be allocated on the memory pool and pinned


MichaelSt98 (approved):
Looks good to me!

PR #104, opened 2025-08-04
Disable F-API memory pool

awnawab:
When enabling GPU offload for more of the IFS, sharing the F-API memory pool across the IFS and ecwam leads to strange offload related errors. Whilst this is investigated, a quick fix is to simply disable the use of the memory pool in ecwam. Fortunately, there is only one temporary F-API field in ecwam, all other fields are allocated once and persist across timesteps. Therefore if this one field can also be made persistent, there is no performance degradation from disabling the memory pool. The only penalty is a small uptick in the host memory footprint.


reuterbal (approved):
Looks good to me!

jkousal32 (approved):

PR #105, opened 2025-08-15
Switch to SCCVec transformation

awnawab:
SCCSeq and SCCVec are the two baseline loki transformations for GPU offload. Although they are both quite slow and would never be used in a proper run, they are a very useful debugging baseline. Of the two, SCCVec is arguably the more useful debugging tool because it doesn't change subroutine interfaces, and therefore if openacc compilation is disabled it can be used effectively to bisect the loki transformed call tree.

This PR switches ecwam to use the SCCVec transformation, bringing it in line with the IFS. SCCSeq also generates unexpected runtime errors for larger grids, these remain to be investigated.


mlange05 (approved):
LGTM 👍

PR #106, opened 2025-08-20
Update deprecated pinning flag

awnawab:
A small PR to update the now deprecated `-gpu=pinned` linker flag for page-locked host allocations with NVHPC.


src/ecwam/CMakeLists.txt
```diff
@@ -483,7 +483,11 @@ foreach( prec sp dp )
     endif()
 
     if( HAVE_CUDA )
```
wdeconinck:
```suggestion
    if( HAVE_CUDA AND CMAKE_Fortran_COMPILER_ID MATCHES NVHPC )
```
Checking for NVHPC compiler just to be sure

PR #107, opened 2025-10-01
update parameters.yml to be in line with mpcrtbl

jkousal32:
very noninvasive, it only affects the standalone running, specifically which codes in parameters.yml correspond to which outputs

### DIFF
--- a/share/ecwam/parameters.yml
+++ b/share/ecwam/parameters.yml
@@ -170,67 +170,76 @@ parameters:
         '057':
                 index: 57
                 description: Spectral skewness
-        sst:
+        '058':
                 index: 58
-                description: Specified sea surface temperature (NEMO)
+                description: Sea ice cover (NEMO)
         '059':
                 index: 59
-                description: Sea ice cover (NEMO)
-        '060':
-                index: 60
                 description: Sea ice thickness (NEMO)
         ucurr:
-                index: 61
+                index: 60
                 description: U-component of current (NEMO)
         vcurr:
-                index: 62
+                index: 61
                 description: V-component of current (NEMO)
+        '062':
+                index: 62
+                description: Wave energy flux magnitude
         '063':
                 index: 63
-                description: Wave energy flux magnitude
+                description: Wave energy flux direction
         '064':
                 index: 64
-                description: Wave energy flux direction
+                description: Significant wave height 10<=T<=12
         '065':
                 index: 65
-                description: Significant wave height 10<=T<=12
+                description: Significant wave height 12<=T<=14
         '066':
                 index: 66
-                description: Significant wave height 12<=T<=14
+                description: Significant wave height 14<=T<=17
         '067':
                 index: 67
-                description: Significant wave height 14<=T<=17
+                description: Significant wave height 17<=T<=21
         '068':
                 index: 68
-                description: Significant wave height 17<=T<=21
+                description: Significant wave height 21<=T<=25
         '069':
                 index: 69
-                description: Significant wave height 21<=T<=25
+                description: Significant wave height 25<=T<=30
         '070':
                 index: 70
-                description: Significant wave height 25<=T<=30
+                description: Wave induced sea level correction
         '071':
                 index: 71
-                description: Wave induced sea level correction
+                description: Spectral width index
         '072':
                 index: 72
-                description: Spectral width index
+                description: Number of freak waves events
         '073':
                 index: 73
-                description: Number of freak waves events
+                description: U-component atmospheric stress
         '074':
                 index: 74
-                description: U-component atmospheric stress
+                description: V-component atmospheric stress
         '075':
                 index: 75
-                description: V-component atmospheric stress
+                description: U-component stress into oceans
         '076':
                 index: 76
-                description: U-component stress into oceans
+                description: V-component stress into oceans
         '077':
                 index: 77
-                description: V-component stress into oceans
+                description: Turbulence energy flux into oceans
         '078':
                 index: 78
-                description: Turbulence energy flux into oceans
+                description: Time domain maximum crest height
+        '079':
+                index: 79
+                description: Time domain maximum wave height
+        '080':
+                index: 80
+                description: Space time maximum crest type
+        '081':
+                index: 81
+                description: Space time maximum wave height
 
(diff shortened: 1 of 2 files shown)


awnawab:
Hi @jkousal32,

Apologies for getting to this so late, and thanks a lot for this cleanup 🙏 Just to double check how to this was tested; you did a standalone run with all output parameters enabled and that completed successfully? We don't have validation hashes for them all, so in this case successful completion should be enough. And you can ignore the failing hpc-ci tests, that's due to an unrelated reason and I'll fix these in a separate PR.

share/ecwam/parameters.yml
jrbidlot, line 1: It looks correct to me.
Now I know that when I had new parameters, I also need to modify this file...

jrbidlot (approved):

wdeconinck:
This PR is being forgotten. ping @awnawab @jkousal32

jkousal32:
Hi @awnawab  @jrbidlot @wdeconinck 
Sorry for the delay
I’ve tried to run a test now with all the output turned on.

A little bit more input required here: 

The following won’t write, but not expected to (experimental):
-	004
-	008

The following don’t work, but should. They can be found here: https://codes.ecmwf.int/grib/param-db/ . I’ve checked I’m using the master version of eccodes. Any ideas what’s going on here?
-	151131
-	151132
-	174098

This one doesn’t exist at all (either by number or name). I think I should delete it.
-	140210

I will delete these (outdated):
-	58
-	59
-	60
-	61

jrbidlot:
I would suggest to not delete any entries but simply ensure that it is consistent with mpcrtbl

140210 was never introduced but we might need it one day with work in sea ice

I think 151131, 151132 and 140098 are not in grib2, at least by default.
We actually already output our data in grib2 (see NGRIB_VERSION in model namelist)

We might need to check that indeed these parameters are not in grib2

jkousal32:
@jrbidlot What do you mean that they are not in grib2? When I look them up in the parameter database, I see the GRIB2 tag there (for 151131, 151132,174098):
Screenshot 2026-01-23 at 09 24 10

jrbidlot:
Back in June, we had the following exchange in Team:
Yes. DGOV in all their wisdom decided that we can not use 151* parameters in GRIB2, ...

[
](https://teams.microsoft.com/l/message/19:21ff209d9e2d497f973feb6214ae6d9d@thread.v2/1749812210063?context=%7B%22contextType%22%3A%22chat%22%7D)

So I thought it might the reason.

Yes, they seem to appear in the parameter database,
but their use might be limiter to S2S data

Screenshot 2026-01-23 at 11 03 14

Best to check what data GOV has allowed in eccodes:
Try with Shahram Najm 

Alternatively, maybe we could try to go for proper grib2 parameters as defined by WMO (see parameter data base):
151131 -> 262140
151132 -> 262139
174098 -> 262000

! note that it ..131 for eastward component and ---132 for northward component but the grib2 ones are ---140 for eastward component and ...139 for northward component

share/ecwam/parameters.yml
```diff
@@ -170,67 +170,76 @@ parameters:
...
+                description: U-component stress into oceans
         '076':
                 index: 76
-                description: U-component stress into oceans
+                description: V-component stress into oceans
         '077':
                 index: 77
-                description: V-component stress into oceans
+                description: Turbulence energy flux into oceans
         '078':
```
awnawab: This should be `tdcmax`.

share/ecwam/parameters.yml
```diff
@@ -170,67 +170,76 @@ parameters:
...
                 index: 78
-                description: Turbulence energy flux into oceans
+                description: Time domain maximum crest height
+        '079':
```
awnawab: Should be `tdhmax`.

share/ecwam/parameters.yml
```diff
@@ -170,67 +170,76 @@ parameters:
...
+                index: 79
+                description: Time domain maximum wave height
+        '080':
```
awnawab: Should be `stcmax`.

share/ecwam/parameters.yml
```diff
@@ -170,67 +170,76 @@ parameters:
...
+                index: 80
+                description: Space time maximum crest type
+        '081':
```
awnawab: Should be `sthmax`.

share/ecwam/parameters.yml
```diff
@@ -170,67 +170,76 @@ parameters:
         '057':
                 index: 57
                 description: Spectral skewness
-        sst:
+        '058':
```
awnawab: This should be `sic`.

share/ecwam/parameters.yml
```diff
@@ -170,67 +170,76 @@ parameters:
...
         '078':
                 index: 78
-                description: Turbulence energy flux into oceans
+                description: Time domain maximum crest height
+        '079':
+                index: 79
+                description: Time domain maximum wave height
+        '080':
+                index: 80
+                description: Space time maximum crest type
```
awnawab: Space time maximum crest *height*.

awnawab (changes requested):
Thanks @jkousal32 for picking this up again 🙏 

For the purposes of this PR, it would be sufficient to simply ensure the field names and descriptions match across MPCRTBL and `parameters.yml`. This is mostly true after your fixes, apart from a few inconsistencies which I've flagged. Could you please address this before we merge this PR?

As for the discussion on which parameters to keep or not. I would say the parameters that have eccodes IDs but generate runtime errors (151131, 151132,174098), we should keep those and discuss the problem and potential solutions with the eccodes team. The parameters that don't have eccodes IDs, these should be deleted (in a future PR, not the current one). If there is a need for them in the future and they obtain an eccodes ID, then we can re-introduce them. The code will stay in the git history and we can restore it easily.

jkousal32:
The following three parameters have been updated to their new grib2 equivalents detailed [here](https://confluence.ecmwf.int/display/MTG2US/Migration+to+GRIB2+-+changes+to+encoding+of+parameters#MigrationtoGRIB2changestoencodingofparameters-Oceanmodeloutput)
-     151131 -> 262140
-     151132 -> 262139
-     174098 -> 262000

@awnawab I have updated all of your requests except for '058' to sic (we have ci as 55)

awnawab (approved):
Thanks a lot @jkousal32 for addressing the comments, this now looks G2G 👌

PR #108, opened 2025-10-07
Avoid global field during allocation

lukasm91:
Currently, we allocate `FIELDG` only to call `INIT_FIELDG`, which itself is used exclusively for computing the `XLON` and `YLAT` arrays. We also deallocate `FIELDG` within the same routine.

Since the calculation of these arrays is straightforward and does not require any global data, this PR proposes to compute them directly. A further improvement could be to move the `XLON`/`YLAT` computation from `INIT_FIELDG` into a dedicated function to avoid duplicating this code.


awnawab:
Hey @lukasm91. Thanks a lot for this, it looks like a very sensible change to reduce memory consumption. The CI is currently fubar'd, once #110 is merged I'd like to rebase this over develop and let the CI run through, but essentially this change looks good 👌

awnawab (approved):

PR #111, opened 2025-10-22
Minimal changes for compatibility with MPL_F08

awnawab:


awnawab:
@ioanhadade FYI

src/ecwam/yownemoio.F90
```diff
@@ -83,7 +84,7 @@ SUBROUTINE WAMININEMOIO(LNEMOIO)
 
 IF (LLNEMOIO) THEN
   CALL NEMOGCMCOUP_INIT_IOSERVER( ILOCAL_COMM, LNEMOIOSERVER )
-  ICOMM=MPI_COMM_WORLD
+  ICOMM=MPI_COMM_WORLD%MPI_VAL
```
wdeconinck: This is not compatible with fiat's MPI_F77_DEPRECATED feature.
I have this PR https://github.com/ecmwf-ifs/fiat/pull/70 in fiat to keep MPL_MPIF intact so that MPI_COMM_WORLD can stay as just INTEGER here, independent from the MPI_F77_DEPRECATED feature.
Let's wait a little while longer to have it merged.

wdeconinck (changes requested):

src/ecwam/yownemoio.F90
awnawab: Thanks a lot for looking into this @wdeconinck and for filing a fix in FIAT 🙏  And yes of course, let's wait for that to be merged.

wdeconinck: Hello @awnawab https://github.com/ecmwf-ifs/fiat/pull/70 is now merged, so that `ICOMM=MPI_COMM_WORLD` can remain as was, provided that you still keep the new introduction of
```Fortran
USE MPL_MPIF, ONLY : MPI_COMM_WORLD
```

awnawab:
I have tested this with unwam enabled, both without and without mpi against FIAT develop. GPU-aware mpi now works again for versions of nvhpc newer than 24.7 thanks to the FIAT changes, so G2G from my side.

awnawab:
The macos-13 runners are EOL, so I've updated those too to have a green CI before we merge.

PR #112, opened 2025-11-07
Switch to sites.ecmwf.int

awnawab:
As nexus is now deprecated, we need to retrieve the input files for ecwam's standalone runners from the correct address.


wdeconinck:
Nice, small change :)

PR #113, opened 2025-11-11
Set generatingProcessIdentifier via namlist

ekuw:
At the moment the generatingProcessIdentifier for the wave model is hardcoded in yowgribhd.F90. With this PR I'd like to switch to instead setting this via namelist variables. 

If accepted, I suggest we propagate this to the main branch for implementation in CY50R1 too, since at the moment the hardcoded values for CY49R2 are being used.

@jrbidlot and/or @jkousal32 it would be great to get your eyes on this.

### DIFF
--- a/src/ecwam/userin.F90
+++ b/src/ecwam/userin.F90
@@ -473,16 +473,28 @@ SUBROUTINE USERIN (IFORCA, LWCUR)
         LFDBIOOUT=.FALSE.
       ENDIF
 
-      IF (IMDLGRBID_M < 0 .OR. IMDLGRBID_G < 0) THEN
+      IF (IMDLGRBID_M < 204 .OR. IMDLGRBID_M > 220) THEN
         WRITE(IU06,*)'++++++++++++++++++++++++++++++++++++++++++++++++++'
         WRITE(IU06,*)'+                                                +'
         WRITE(IU06,*)'+ SUBROUTINE USERIN :                            +'
         WRITE(IU06,*)'+ READ NAMELIST FAILED                           +'
-        WRITE(IU06,*)'+ IMDLGRBID_M and IMDLGRBID_G MUST BE SPECIFIED  +'
+        WRITE(IU06,*)'+ IMDLGRBID_M MUST BE IN THE RANGE 204 - 220     +'
         WRITE(IU06,*)'+ PROGRAM WILL ABORT                             +'
         WRITE(IU06,*)'+                                                +'
         WRITE(IU06,*)'++++++++++++++++++++++++++++++++++++++++++++++++++'
-        CALL WAM_ABORT("Expected namelist input for IMDLGRBID_M and IMDLGRBID_G",__FILENAME__,__LINE__)
+        CALL WAM_ABORT("Unexpected namelist input for IMDLGRBID_M",__FILENAME__,__LINE__)
+      ENDIF
+
+      IF (IMDLGRBID_G < 104 .OR. IMDLGRBID_G > 120) THEN
+        WRITE(IU06,*)'++++++++++++++++++++++++++++++++++++++++++++++++++'
+        WRITE(IU06,*)'+                                                +'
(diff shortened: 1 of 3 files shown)


ekuw:
I made some updates based on feedback from @jrbidlot to remove the now redundant initialisation of these model numbers in yowgribhd and to move the documentation around these variables into mpuserin.

jkousal32 (approved):
Looks good to me!

src/ecwam/userin.F90
jrbidlot, line 1: Sorry, I have just thought of it.
The rule is for  104 <= IMDLGRBID_G <= 120
and 204 <= IMDLGRB_M <= 220
I suggest that we check for that.

src/ecwam/yowgribhd.F90
jrbidlot, line 1: OK, there as a reason why  these parameters were hardcoded because they are used by    preset_wgrib_template which is called by the main programme but also by the few other utilities:
   preproc, preset, intwaminput and create_wam_bathymetry_ETOPO1.F90
     We could try the add these parameters to the respective input of those utilities but it would complicate the scripts a lot for not reason as these utilities will not produce data that would be disseminated.
     Instead, I suggest to give the 2 variables a default value that is within the expected definition:
     
      INTEGER(KIND=JWIM) :: IMDLGRBID_G=108 
      INTEGER(KIND=JWIM), :: IMDLGRBID_M=208

jrbidlot (changes requested):
Welcome the world of scripts duplication:
ecwam has his own scripts, therefore any change made to the main wave model namelist in the ifs-scripts branch need to be reflected in the ecwam github:
See
/share/ecwam/scripts/ecwam_run_model.sh
which contain the main namelist that now needs to new entries
108 and 208 respectively
Because these scripts are only used for testing, it does not really matter which values we use as long as that they follow the rule:  104 <=.   <=120 and 204 <=   <= 220

src/ecwam/yowgribhd.F90
awnawab: I agree, it seems like a cleaner solution to have the values be initialised to a sensible default in `YOWGRIBHD`, and then be overridable in `MPUSERIN` via namelist.

src/ecwam/userin.F90
awnawab: Yes, we can simply hardcode a sensible value to `ecwam_run_model.sh`, rather than try to make it configurable via the yaml test configs like some of the other namelist inputs.

awnawab:
Thanks a lot @ekuw for this contribution 🙏 Please ignore the failing tests for now, I will patch develop-1.4 shortly to restore the CI and then I'll rebase your branch over it so that we can test it before merging.

awnawab:
Hi @ekuw,

CI fixes have now been merged to develop-1.4, can I please rebase your branch over develop-1.4 so that we can run the CI?

ekuw:
Hi @jrbidlot and @awnawab - thanks a lot for your comments. I've been through them and made an update based on my understanding of your discussion. This includes:

- changing the initial values in mpuserin and yowgribhd from to 108/208 for IMDLGRBID_G/M
- updating the check in userin to check for valid values of IMDLGRBID_G/M
- rebasing on develop-1.4

I would say that in an ideal world we would update the scripts that require these values so that we *require* them in the namelist. As it is now, if someone doesn't specify these in the namelist then they automatically (and silently) pick up the potentially incorrect default. Maybe this should be an improvement to think about for later?

In any case, please let me know if what I've done is inline with your expectations - thanks!!

jrbidlot:
Hi Emma, when used at ECMWF the current scripts would provide necessary information for grib encoding
so I am so worried about hardcoding values for IMDLGRBID_G/M
For outside use, these values are really not relevant as they are specific to ECMWF definition.

jrbidlot (approved):
It should be good to go.

awnawab (approved):
Thanks a lot @ekuw for addressing the suggestions 🙏 this is now G2G 👌

PR #116, opened 2025-11-18
Fix python search

awnawab:
A small fix to the python search behaviour, which removes a safeguard only needed for projects that create python virtual environments so is not needed here. This allows for example, building ecWAM by having an active virtual env that has fypp and pyyaml installed.


wdeconinck (approved):
Thanks @awnawab that works for me!

PR #117, opened 2025-11-20
Fix warning on conversion with -r8 flag

wdeconinck:
The NVHPC Fortran compiler warning:

```
NVFORTRAN-W-0155-The type of FLOAT is now double precision with -r8 (propconnect.F90: 686)
```


jrbidlot (approved):
That float must have escape. We should indeed use real( ,jwrb)

awnawab (approved):
Thanks a lot for this fix @wdeconinck 🙏 🙏

PR #118, opened 2025-11-20
Avoid zero-sized allocation in readmdlconf

wdeconinck:
In READMDLCONF we had situation that ...

fields were allocated first with size NIBLO (original line 76) which is apparently zero in standalone runs.
Then immediately fields are deallocated again on line 109 and allocated again on line 110 with an updated size NIBLO.

This allocation/deallocation/allocation can simply be avoided with this PR.

Now one extra thing which I would want @jrbidlot and @jkousal32 's opinion on is that it seems that in the case of `LLREAD=.FALSE.` (the ELSE of the IF statement),  NIBLO is computed differently as NGX*NGY, but the fields were not reallocated with this new computed size and stayed with the original NIBLO=0 size allocated on line 76.
In this PR I also try to fix this but I am not sure this was a bug or intended. If intended I'd have to allocate it to replicate the original behaviour for `LLREAD=.FALSE.`.

Perhaps the original NIBLO used in line 76 is non-zero in an IFS-coupled configuration?


awnawab (approved):
Regardless of what the value of `NIBLO` should be, which Jean and Josh can better comment on, the new logic seems correct to me 👌

jrbidlot:
Thank you Willem for pointing to this issue.
The call to READMDLCONF in the main part of code is without any of the optional arguments.
They are however used in the programme intwaminput, currently in ifs project wam/Wam_oper/intwaminput.F90
There are two calls to it which both go the the else condition. But indeed for each one of them NIBLO should be NIBLO=NGX*NGY (because it is taking the full size grid and does not need to know the total number of ocean points)
Your changes should work.

jrbidlot (approved):

PR #119, opened 2025-11-21
Avoid use of uninitialised allocations

wdeconinck:
With alternative field_api allocation mechanism and in a DEBUG build, allocated data gets initialised to SNAN.
This showed that uninitialised data was being used through a "INVALID" FPE being signalled.
The safest way is to not rely on compiler flags or allocators but to set values to zero manually. This is now done within the `_ALLOC` functions of the generated derived types.

This solution to was debugged and developed together with @awnawab, Thanks!


awnawab (approved):
Thanks a lot for another cleanup PR 🙏 🙏  G2G

PR #122, opened 2025-11-27
Various offload fixes and improvements

awnawab:
@wdeconinck reported a set of fields which were being implicitly copied to device. Rather than relying on this behaviour, we should make this data movement explicit.

This PR addresses the above bug and also takes the opportunity to remove a lot of redundant data transfers, resulting in an 85% reduction in data transfer times in an O320 DP run on a single Atos node. The updated data-transfer logic has also been tested successfully in a coupled IFS build with ecwam on GPU and the atmosphere on CPU (performance testing in IFS coupled runs is still pending).


wdeconinck:
I can confirm that my draft PR atlas_field_api rebased on this branch works without the commit I needed to add to fix the offload without pinning. It works with and without pinning.

wdeconinck (approved):

PR #123, opened 2026-01-22
Replace use of local cmake macro ecwam_target_fortran_module_directory with ecbuild

wdeconinck:
Just a general code maintenance cleanup.

The ecbuild macro was added in ecbuild 3.8, released ~2.5 years ago.
So this PR raises minimum required version to 3.8.

This allows to remove the ecwam macro.


awnawab (approved):
Many thanks for this cleanup @wdeconinck, looks great 👌

PR #126, opened 2026-02-17
Plume - WAM entry point (expose swh only)

cducher:
This PR adds an entry point for Plume in WAM. The main Plume hook is in IFS, and it intialises the callbacks to the Plume API (see associated PRs in IFS), therefore, Plume cannot be used in WAM standalone runs at the moment.
Plume exposes a selection of WAM fields to its plugins, it was demonstrated here with the swh only, as per discussions with @jkousal32 and @jrbidlot, please advise if another way should be favoured.


src/ecwam/wam_plume_module.F90
```diff
@@ -0,0 +1,99 @@
...
+  REAL(KIND=JWRB), DIMENSION(1:NPROMA_WAM,NANG,NFRE) :: FL2ND
+
+  IF (LHOOK) CALL DR_HOOK('COMPUTE_PLUME_FLDS',0,ZHOOK_HANDLE)
+
+!$OMP PARALLEL DO SCHEDULE(DYNAMIC,1) PRIVATE(ICHNK)
+  DO ICHNK = 1, NCHNK
+    FL2ND(1:NPROMA_WAM,:,:) = SPECTRA(1:NPROMA_WAM,:,:, ICHNK)
+    CALL FEMEAN (1, NPROMA_WAM, FL2ND, EM, FM)
+    ! SIGNIFICANT WAVE HEIGHT CONVERSION
+    PLUME_FLDS(1:NPROMA_WAM, 0, ICHNK)=4._JWRB*SQRT(MAX(EM(1:NPROMA_WAM),0._JWRB))
```
jrbidlot: Is this correct?
Should it be
PLUME_FLDS(1:NPROMA_WAM, **1,** ICHNK)=4._JWRB*SQRT(MAX(EM(1:NPROMA_WAM),0._JWRB))

src/ecwam/wam_plume_module.F90
```diff
@@ -0,0 +1,99 @@
...
+  REAL(KIND=JWRB), DIMENSION(NPROMA_WAM, 1, NCHNK) :: PLUME_FLDS
+  INTEGER(KIND=JWIM) :: ICHNK
+  REAL(KIND=JWRB), DIMENSION(1:NPROMA_WAM) :: EM, FM
+  REAL(KIND=JWRB), DIMENSION(1:NPROMA_WAM,NANG,NFRE) :: FL2ND
+
+  IF (LHOOK) CALL DR_HOOK('COMPUTE_PLUME_FLDS',0,ZHOOK_HANDLE)
+
+!$OMP PARALLEL DO SCHEDULE(DYNAMIC,1) PRIVATE(ICHNK)
+  DO ICHNK = 1, NCHNK
+    FL2ND(1:NPROMA_WAM,:,:) = SPECTRA(1:NPROMA_WAM,:,:, ICHNK)
```
jrbidlot, line 90: If you check outblock.F90, you will see that the variable FL2ND was introduced to be a copy to SPECTRA (called FL1 in outblock) only if no refraction: IREFRA == 2 .OR. IREFRA == 3 or no second order correction (LSECONDORDER false).
For now, since PLUME will only output the significant wave height, it will be fine as it is because both transformations (INTPOL and CAL_SECOND_ORDER_SPEC) conserve wave energy and  hence also the significant wave height. But we should add a comment that FL2ND is the second order spectrum in the absolute frame of reference (see outblock)

Relevant lines in outblock
      IF (IREFRA == 2 .OR. IREFRA == 3) THEN
        CALL INTPOL (KIJS, KIJL, FL1, FL2ND, WAVNUM, UCUR, VCUR, IRA)
      ELSE
        FL2ND(KIJS:KIJL,:,:) = FL1(KIJS:KIJL,:,:)
      ENDIF
      IF (LSECONDORDER) CALL CAL_SECOND_ORDER_SPEC(KIJS, KIJL, FL2ND, WAVNUM, DEPTH, SIG)

src/ecwam/wam_plume_module.F90
```diff
@@ -0,0 +1,99 @@
...
+
+PROCEDURE(WAM_PLUGINS_RUN), POINTER :: WAM_PLUME_HANDLER_RUN_HANDLER => NULL()
+PROCEDURE(WAM_IFS_PLUGINS_RUN), POINTER :: WAM_IFS_PLUME_HANDLER_RUN_HANDLER => NULL()
+
+CONTAINS
+
+! Function pointer to call the run method on the Plume instance from WAM sub time steps
+SUBROUTINE WAM_PLUGINS_RUN(PLUME_FLDS, TIME_DIFF, WSTEP_UPDATED)
+  REAL(KIND=JWRB), DIMENSION(NPROMA_WAM, 1, NCHNK), INTENT(IN) :: PLUME_FLDS
+  INTEGER(KIND=JWIM), INTENT(IN) :: TIME_DIFF
```
jrbidlot: Ideally, integer variables should start with I, J, or N

src/ecwam/wam_plume_module.F90
```diff
@@ -0,0 +1,99 @@
...
+  LOGICAL, INTENT(IN) :: WSTEP_UPDATED
```
jrbidlot: Ideally, logical variables should start with LL

src/ecwam/wam_plume_module.F90
```diff
@@ -0,0 +1,99 @@
...
+  REAL(KIND=JWRB), DIMENSION(NPROMA_WAM, NANG, NFRE, NCHNK), INTENT(IN) :: SPECTRA
+  REAL(KIND=JPHOOK) :: ZHOOK_HANDLE
+  REAL(KIND=JWRB), DIMENSION(NPROMA_WAM, 1, NCHNK) :: PLUME_FLDS
+  INTEGER(KIND=JWIM) :: ICHNK
+  REAL(KIND=JWRB), DIMENSION(1:NPROMA_WAM) :: EM, FM
+  REAL(KIND=JWRB), DIMENSION(1:NPROMA_WAM,NANG,NFRE) :: FL2ND
+
+  IF (LHOOK) CALL DR_HOOK('COMPUTE_PLUME_FLDS',0,ZHOOK_HANDLE)
+
+!$OMP PARALLEL DO SCHEDULE(DYNAMIC,1) PRIVATE(ICHNK)
```
jrbidlot: FL2ND should also be private

jrbidlot (changes requested):

CMakeLists.txt
```diff
@@ -54,6 +54,10 @@ ecbuild_add_option( FEATURE UNWAM
                     DEFAULT OFF
                     DESCRIPTION "Support for UNWAM" )
 
+ecbuild_add_option( FEATURE PLUME
+                    DEFAULT OFF
```
awnawab, line 58: Might be a good idea to add `REQUIRED_PACKAGES "plume"` here as an extra safety measure to ensure everything is set up properly.

src/ecwam/wam_plume_module.F90
```diff
@@ -0,0 +1,99 @@
...
+#ifdef WITH_PLUME
+
+#define __FILENAME__ "wam_plume_module.F90"
+USE PARKIND_WAVE, ONLY : JWRB, JWIM
+USE YOWGRID,      ONLY : NPROMA_WAM, NCHNK
+USE YOWPARAM,     ONLY : NANG, NFRE
+USE YOMHOOK,      ONLY : LHOOK, DR_HOOK, JPHOOK
+USE YOWABORT,     ONLY : WAM_ABORT
+
+IMPLICIT NONE
```
awnawab, line 26: Could you please add a `PRIVATE` here to ensure only the symbols explicitly marked as public are exported?

src/ecwam/wam_plume_module.F90
awnawab: As well as EM, FM

src/ecwam/wam_plume_module.F90
awnawab: I agree, unless explicitly specified array dimensions have a lower bound of 1 by default, so the middle index should indeed be 1.

src/ecwam/wam_plume_module.F90
```diff
@@ -0,0 +1,99 @@
...
+!-----------------------------------------------------------------------
+MODULE WAM_PLUME_MODULE
+!
+!== wam_plume_module.F90 ==
+!
+! A module to encapsulate Plume-calls from WAM. Not intended for standalone runs for now.
+!
+! Author: Clara Ducher, 11-Feb-2026
+!
+#ifdef WITH_PLUME
```
awnawab: Considering the import of this module is ifdef guarded, and all of the symbols it defines are ifdef guarded, there's no point compiling it if plume is disabled. So I would remove this ifdef here, and instead conditionally add `wam_plume_module.F90` to ecwam_srcs only if `HAVE_PLUME` is true.

awnawab (changes requested):
Thanks @cducher for this contribution 🙏  you've very neatly implemented a complex feature, well done! I have just a few minor comments, on top of @jrbidlot's comments, that should be addressed before this is merged. Could you also please share a link to the matching ifs PR? It would be nice to have a bit more context about this development. Thanks! you can ignore the one failing CI test, that's an unrelated bug that I'll look into to.

cducher:
Hi both, thank you for your prompt feedback, I will look into addressing your comments once I am back from holidays in March. In the meantime @awnawab, please feel free to also have a look at the IFS PRs for the full update:

- ifs-bundle (update the ecwam version to fetch the changes proposed here): https://git.ecmwf.int/users/maek/repos/ifs-bundle/pull-requests/54/overview
- ifs-source (add callback mechanism for a Plume - WAM entry point): https://git.ecmwf.int/users/maek/repos/ifs-source/pull-requests/87/overview
- ifs-test (modify Plume test to run with WAM coupling): https://git.ecmwf.int/users/maek/repos/ifs-test/pull-requests/17/overview

jkousal32 (approved):

awnawab:
Thanks a lot @cducher for addressing the comments 🙏 the new CI failures are related to a quirk of fcm, if you replace the unqualified `WAM_PLUME_MODULE` import with a qualified one importing only those symbols used in `WAMODEL`, then the error will go away.

jrbidlot (approved):
As far as I can tell. It looks OK.

awnawab (approved):

PR #128, opened 2026-03-01
hpc-ci: use ecwam-bundle

awnawab:
For security reasons, a pull_request_target event runs on the repository's default branch, which in the current case is develop. This means that regardless of the PR target branch, the hpc-ci will run the state of the workflow in develop.

Currently, the hpc-ci builds ecWAM's dependencies independently rather than using a bundle build, and the version of these dependencies is therefore encoded within the workflow itself. This is potentially problematic if we have to support multiple cycles, and each cycle may require different versions of ecWAM's dependencies. Using a bundle build in the hpc-ci avoids this problem, as now the dependencies are checked out at the versions contained in the bundle.

This PR updates the hpc-ci to use a bundle build.

### DIFF
--- a/.github/workflows/build-hpc.yml
+++ b/.github/workflows/build-hpc.yml
@@ -42,16 +42,9 @@ jobs:
               #SBATCH --gpus-per-task=1
               #SBATCH --mem=0
               #SBATCH --qos=dg
-            modules:
-              - cmake
-              - fcm
-              - ninja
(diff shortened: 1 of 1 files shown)


jkousal32 (approved):
Nice one! Looks good to me

wdeconinck (approved):
Makes sense!

PR #134, opened 2026-04-10
Harmonise `WVALLOC` and `WVDEALLOC`

awnawab:
This PR harmonises `WVALLOC` and `WVDEALLOC` so that the latter deallocates only what the former allocates. Grid related quantities, which are expensive to compute and we want to persist during a multi trajectory oops run, are therefore left allocated. This does not carry any risk of a memory leak, it just means three ecwam data-structures are deallocated implicitly at the end of the program execution.


awnawab:
I'm leaving the allocation of `F_BOUT` in `WAMODEL` since it `NIPRMOUT` isn't initialised by the time we call `WVALLOC`.

jrbidlot (approved):
OK.
I now understand how F_BOUT is handled but now that we have made the statement that wvdealloc mirrors wvalloc, should we add a comment that it is not true for F_BOUT?

prgillies (approved):
Thanks for this!

jrbidlot (approved):
Thank you.

jkousal32 (approved):

awnawab:
Thanks @prgillies for confirming that this solves the problem you were hitting in the 50R3 CI, merging now.

PR #135, opened 2026-04-14
Support multi precision nemo builds

awnawab:
This PR makes the necessary changes to support multiple precision nemo builds: https://github.com/ecmwf-ifs/ifs-nemo/pull/5

I've also taken this opportunity to clean up some leftover code from when ecwam also gained support for multi precision builds.

Please note that because this is a non backwards compatible change, this PR should only be merged once all associated PRs have been approved. Their status can be tracked on this FORGE ticket: https://jira.ecmwf.int/browse/FORGE-333


jrbidlot (approved):

awnawab:
Thanks @jrbidlot for the review 🙏  @wdeconinck could you also please give this a quick look? The corresponding ifs-bundle and ifs-nemo PRs were already approved.

jkousal32 (approved):

wdeconinck (approved):

awnawab:
@wdeconinck as discussed offline, I've adapted this to make the changes backwards compatible with 50R1. Please have another look when you can 😊

CMakeLists.txt
```diff
@@ -55,8 +55,12 @@ set( HAVE_sp ${HAVE_SINGLE_PRECISION} )
...
 ecbuild_add_option( FEATURE OCEAN_COUPLING
                     DEFAULT ON
-                    DESCRIPTION "Support ocean coupling"
-                    REQUIRED_PACKAGES "nemo" )
+                    DESCRIPTION "Support ocean coupling")
+### Determine ocean model precision
+if( HAVE_OCEAN_COUPLING )
+  include( ecwam_find_nemo )
+endif()
+
```
wdeconinck, line 62: This workflow makes NEMO by default enabled, and ecwam_find_nemo will throw critical error when NEMO not found. Any workflow without NEMO will stop working, unless `ENABLE_OCEAN_COUPLING=OFF` is set explicitly.

The REQUIRED_PACKAGES keyword in `ecbuild_add_option` was doing all the heavy work here.
You could modify the hard abort with a check if the user specified ENABLE_OCEAN_COUPLING:
```cmake
if (DEFINED ${PNAME}_ENABLE_OCEAN_COUPLING)
  if (${PNAME}_ENABLE_OCEAN_COUPLING)
      ... NEMO MUST BE FOUND OR ABORT ...
  endif()
else if(DEFINED ENABLE_OCEAN_COUPLING)
  if (ENABLE_OCEAN_COUPLING)
    ... NEMO MUST BE FOUND OR ABORT ...
  endif()
else()
   ... User did not explicitly request OCEAN_COUPLING, so it's OK if not found ...
   ... Just report feature OFF ...
    set(HAVE_OCEAN_COUPLING 0)
endif()
```
and then there is the issue that the feature summary will not be displaying the feature correctly as ON or OFF...

wdeconinck (changes requested):

CMakeLists.txt
awnawab: Hmm yes good point. I'm surprised the CI still works somehow. What if I were to move ecwam_find_nemo before the option and make that a condition for the option? This would also fix the summary problem you highlighted.

wdeconinck: Yes indeed can also be.

wdeconinck (approved):
Looks good now. I trust you tested the backward compatibility with an old setup as well as the new setup.

awnawab:
Thanks @wdeconinck 🙏 Yes I successfully built all four variants of ifs/nemo precision combinations for both 50r1 and 50r3.

PR #137, opened 2026-04-28
Update GNU 15.2.0 flags

ioanhadade:
### Description
Updated flags for GNU 15.2.0 which are needed also to prevent a SIGFPE in sdice3.

### Contributor Declaration


jrbidlot (commented):
Do we have any idea which line is causing the SIGFPE?

If so, is the only way is to compile it differently rather than adjusting the code?

ioanhadade:
The SIGFPE was raised because GNU 15.2.0 was substituting the call to POW with a vectorized version that uses a different ULP. No code has been changed, just the flag for that particular file which prevents the GNU compiler for doing that aggressive optimization (i.e., substitute POW with a SIMD/vectorized clone).

src/ecwam/CMakeLists.txt
```diff
@@ -504,17 +481,14 @@ endforeach()
 if( CMAKE_Fortran_COMPILER_ID MATCHES Intel )
   set_source_files_properties( mubuf.F90 PROPERTIES COMPILE_OPTIONS "-fp-model;strict" )
   set_source_files_properties( propconnect.F90 PROPERTIES COMPILE_OPTIONS "-fp-model;strict" )
-  set_source_files_properties( stresso.F90 PROPERTIES COMPILE_OPTIONS "-fp-model;strict" )
 elseif( CMAKE_Fortran_COMPILER_ID MATCHES GNU )
   set_source_files_properties( mubuf.F90 PROPERTIES COMPILE_OPTIONS "-ffp-contract=off" )
```
awnawab: Can we now remove this please? Because we are setting `-ffp-contract=off` across the whole project anyway.

awnawab (approved):
Thanks a lot @ioanhadade 🙏  Since we disable fma for other compiler toolchains (for good reason), it's only right to do it for gnu too. I'll have a similar change in 50R3/2 as part of the ecland extraction.

I've left one optional comment, but even as it is this is merge ready. Ignore the failing nvhpc tests, I need to upgrade the CI to a newer nvhpc version.

src/ecwam/CMakeLists.txt
ioanhadade: Personally yes, I think we should.

awnawab:
Sorry @ioanhadade I labelled the wrong line. We need to remove `ffp-contract=off` for mubuf not the flag for stresso. Apologies 😅

awnawab:
Ah you deleted the right line, just the commit message is wrong. Ok nvm, excuse the noise 😄

ioanhadade:
Sorry, yes, commit message got polluted thanks to you but ended up deleting the right line :D

src/ecwam/CMakeLists.txt
```diff
@@ -506,7 +506,7 @@ if( CMAKE_Fortran_COMPILER_ID MATCHES Intel )
   set_source_files_properties( propconnect.F90 PROPERTIES COMPILE_OPTIONS "-fp-model;strict" )
   set_source_files_properties( stresso.F90 PROPERTIES COMPILE_OPTIONS "-fp-model;strict" )
 elseif( CMAKE_Fortran_COMPILER_ID MATCHES GNU )
-  set_source_files_properties( mubuf.F90 PROPERTIES COMPILE_OPTIONS "-ffp-contract=off" )
+  set_source_files_properties( sdice3.F90 PROPERTIES COMPILE_OPTIONS "-fdisable-ipa-simdclone" )
```
wdeconinck, line 509: Is this due to a compiler bug? If so, can this be reproduced by someone and reported?

ioanhadade:
It is related to this: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=118955

wdeconinck:
> It is related to this: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=118955

Thanks @ioanhadade for this. So it is a **glibc** bug seemingly, and likely that GCC 16 will have the fix for this then. Should we perhaps add that in a comment for future reference? So that we know when to remove such flag?

awnawab:
> > It is related to this: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=118955
> 
> Thanks @ioanhadade for this. So it is a **glibc** bug seemingly, and likely that GCC 16 will have the fix for this then. Should we perhaps add that in a comment for future reference? So that we know when to remove such flag?

I'm preparing a pr to fix the nvhpc CI issues, so I'll add a comment with a link to that bug report. Once gcc 16 is released we can test it and remove that workaround. So I'll merge this is as is. Thanks all for the discussion 🙏

src/ecwam/CMakeLists.txt
jrbidlot: @ioanhadade 
for my own understanding, is this disabling of the aggressive optimisation applied only to sdice.F90 ?

If so, sdice3 does not have COS()  ?

ioanhadade: it's only applied to sdice3.F90 because that is the only place at least that raised a SIGFPE. If we see more places where there is a cos() or trig function called with GNU compiler, using the above flag will likely fix the issue.

jrbidlot: OK. if it does the trick fine but I am still weirded out because sdice3 does not contain trig function...

ioanhadade: I did some further digging, it looks like it is the vectorized powf function generated by the compiler that is the problem. The issue is the same as with cosf and disabling the vectorized clone fixes the problem here.

jrbidlot:
Ioan,
that makes a lot more sense. The parameterization in sdice3 has indeed
power calculation with non-integer exponent.

Thank you for the clarification.

Jean

PR #139, opened 2026-04-29
Update nvhpc CI runners to 25.9

awnawab:
This PR upgrades the nvhpc CI runners to version 25.9.


wdeconinck (approved):
All good. Thanks for adding all the useful comments

PR #141, opened 2026-04-30
logic switch added in mpdecomp.F90 for restart redistribute transformation

antonl321:
### Description

When redistributing the restart files to different number of MPI ranks
nproma_wam must not change.
I added  LLNO_WAM_NPROMA optional argument to MPDECOMP to signal this case
used in ifs-source/wam/Wam_oper/write_mpdecomp.F90 (cy49r3)

### Contributor Declaration


awnawab (approved):
Hi Lucian. Thanks for this change, it looks good to me 🙏 So the idea is that if `LLNO_WAM_NPROMA` is present and true, then we just use whatever `NPROMA_WAM` is specified via namelist? And would you also like this to be included in 50R2? If so I can cherry-pick this change to ecwam develop.

antonl321:
Hi Ahmad,

Yes, that's the idea. This is needed for the program that redistributes  wam restart files to a different number of MPI ranks.
If I remember correctly Balthasar told me that he wants restarts redistribution in CY50 too, so I think that this feature must keep up with it.

PR #151, opened 2026-07-15
Small CI cleanup

awnawab:
This PR applies the following cleanup to the CI:

- Change build type from Release to Bit, bringing it inline with the default build type in bundle builds and ifs coupled builds
- Removes the CI entry for GNU 10, which by this point is a very outdated compiler

Future PRs will also add support for Intel ifx and flang to the CI.


wdeconinck (approved):
That looks good to me!

PR #152, opened 2026-07-17
Ecwam develop fix tests

jrbidlot:
### Description
In the current version of ecWAM, the basic tests available are not up to date with what we do in 50R1:
Namely, it does not include wave in sea-ice and wave current interaction.
Also, all configurations are for a global setup.

I have updated the tests and added a limited area configuration and some configurations that would start from grib restart files rather than pure binary ones (we use grib restart in operations). I have added the test CMakeLists.txt to pickup some of these new configurations (I have not added those that require a bit more resources).

During the testing, I also came across a significant performance deterioration when reading the restart files, which I have solved (bit identical changes) (readfl and readstress).

I have also alter the default namelist in mpuserin to be more inline with 50R1 (in case someone would run without updating their namelist and added a few entries at scripts level to make them more visible following discussion with MétéoFrance who were unaware that they were running with those options.

I would like to have these changes into a new release as it would be more in line with what we do in operations.
I know that with respect to the default tests contained in the branch, it is not bit identical with what was there before and future we will need to make sure we update these tests inline with what we develop when working with the IFS.

### Contributor Declaration


tests/CMakeLists.txt
```diff
@@ -34,13 +38,9 @@ foreach( prec sp dp )
...
     endif()
 
-    ecwam_add_test( ${PROJECT_NAME}_test_etopo1_oper_an_fc_O48_cy49r1 CONFIG etopo1_oper_an_fc_O48_cy49r1.yml PREC ${prec} )
-    ecwam_add_test( ${PROJECT_NAME}_test_etopo1_oper_an_fc_O48_cy49r1 CONFIG etopo1_oper_an_fc_O48_cy49r1.yml MPI 4 OMP 1 PREC ${prec} )
-    ecwam_add_test( ${PROJECT_NAME}_test_etopo1_oper_an_fc_O48_cy49r1 CONFIG etopo1_oper_an_fc_O48_cy49r1.yml MPI 1 OMP 4 PREC ${prec} )
-
-    ecwam_add_test( ${PROJECT_NAME}_test_etopo1_oper_an_fc_O48_cy50r1 CONFIG etopo1_oper_an_fc_O48_cy50r1.yml PREC ${prec} )
-    ecwam_add_test( ${PROJECT_NAME}_test_etopo1_oper_an_fc_O48_cy50r1 CONFIG etopo1_oper_an_fc_O48_cy50r1.yml MPI 4 OMP 1 PREC ${prec} )
-    ecwam_add_test( ${PROJECT_NAME}_test_etopo1_oper_an_fc_O48_cy50r1 CONFIG etopo1_oper_an_fc_O48_cy50r1.yml MPI 1 OMP 4 PREC ${prec} )
+    ecwam_add_test( ${PROJECT_NAME}_test_etopo1_oper_an_fc_grib_restart_O320 CONFIG etopo1_oper_an_fc_grib_restart_O320.yml PREC ${prec} )
```
awnawab: Hi @jrbidlot. I haven't looked through your PR in detail, but I think the problem arrises from adding these tests. The github runners have limited resources, and they aren't suitable for the O320 grids.

It does make sense to also run larger tests in the CI. We can enable this via an environment variable based trigger, for example, and thereby we can run them only for the hpc-ci, which runs on atos nodes.

I can file a PR to develop that sets up this mechanism with the existing O320 test and you can then add your new tests to that same mechanism.

Would it be ok to do this in one week's time? I'm off for the next week, and will be moving lots of boxes and furniture so laptop access will be limited 😅

The other option would be to restrict the current PR to the O48 tests, and introduce the O320 tests to the CI in a later PR.

jrbidlot: Hi Ahmad,
Good luck with you move. The PR can wait until you are fully functional.
I have a few higher resolution configurations that would be good to add the the hpc-ci.
In the meantime, I have revised the low resolution test to include one at O96. It is small enough to run but contains the interaction between waves and ocean currents (part of 50R1). O48 is way too low resolution to have meaningful wave-current interaction, hence why I have added a O96 case.

jrbidlot:
Sorry,
I should have gone through co-pilot review first and then ask for the human review afterwards.
I think it is ready now.

jkousal32 (approved):
Looks good, thanks a lot for that work @jrbidlot

PR #154, opened 2026-07-19
Ecwam develop fix tests

jrbidlot:
### Contributor Declaration


jkousal32:
Is this a duplicate of https://github.com/ecmwf-ifs/ecwam/pull/152 ? If so, maybe delete one of them and then we can work on the remaining one? @jrbidlot

awnawab:
Hi @jrbidlot. Thanks for restricting the CI to O96 grids and below. How did you update the reference norms for the various tests? The convention so far has been to use an ifort debug build as the reference. Is that how you computed them? If not, would you mind if I updated the reference norms using that method and pushing an update to this branch?

jrbidlot:
> Hi @jrbidlot. Thanks for restricting the CI to O96 grids and below. How did you update the reference norms for the various tests? The convention so far has been to use an ifort debug build as the reference. Is that how you computed them? If not, would you mind if I updated the reference norms using that method and pushing an update to this branch?

Well, I did not know of the convention about the norms. I took them from running the tests after after compiling:
sbuild ./package/bundle/ecwam-bundle build -j 32 --arch=package/bundle/arch/ecmwf/hpc2020/intel/2021.4.0/hpcx-openmpi/2.9.0 --with-fckit

I don't mind you updating them
BUT, I have discovered yesterday a very annoying feature.
When running the preproc task, it determines whether it needs to recompute the create_bathymetry part of the job by looking for the files create_bathymetry would create. The files have a md5 key attached to their names.
It is a nice feature as the process can otherwise take a bit of time...
But, it actually does not check if something might have changed in the fortran code create_wam_bathymetry_ETOPO1.F90
Because, I have used the O48 configuration for a while, I have a quite old version of those files.
Since that time, the files containing the subgrid informations have moved from being encoded in grib1 to grib2.
There might also be other differences.
Now that I know, I can go and remove all these old files, but it will not be obvious to any other users...

Short of having the md5 key also connected to all the relevant fortran code, I don' know how to fully avoid that trap.
There is a partial solution built in the script. In ecwam_runtime.sh, the variable ecwam_bathymetry_version=2 could be updated to 3, forcing the recalculation for this new release
What do you think?

wdeconinck:
> Short of having the md5 key also connected to all the relevant fortran code, I don' know how to fully avoid that trap.
> There is a partial solution built in the script. In ecwam_runtime.sh, the variable ecwam_bathymetry_version=2 could be updated to 3, forcing the recalculation for this new release
> What do you think?

Hi @jrbidlot indeed it is the `ecwam_bathymetry_version` variable that should be updated. Fortran code and even ecwam versions can change without having the file needing to be recalculated. If this was not clear enough then it would be best to make this more clear in a README and in comments in the Fortran code.

awnawab:
> > Hi @jrbidlot. Thanks for restricting the CI to O96 grids and below. How did you update the reference norms for the various tests? The convention so far has been to use an ifort debug build as the reference. Is that how you computed them? If not, would you mind if I updated the reference norms using that method and pushing an update to this branch?
> 
> Well, I did not know of the convention about the norms. I took them from running the tests after after compiling: sbuild ./package/bundle/ecwam-bundle build -j 32 --arch=package/bundle/arch/ecmwf/hpc2020/intel/2021.4.0/hpcx-openmpi/2.9.0 --with-fckit
> 
> I don't mind you updating them BUT, I have discovered yesterday a very annoying feature. When running the preproc task, it determines whether it needs to recompute the create_bathymetry part of the job by looking for the files create_bathymetry would create. The files have a md5 key attached to their names. It is a nice feature as the process can otherwise take a bit of time... But, it actually does not check if something might have changed in the fortran code create_wam_bathymetry_ETOPO1.F90 Because, I have used the O48 configuration for a while, I have a quite old version of those files. Since that time, the files containing the subgrid informations have moved from being encoded in grib1 to grib2. There might also be other differences. Now that I know, I can go and remove all these old files, but it will not be obvious to any other users...
> 
> Short of having the md5 key also connected to all the relevant fortran code, I don' know how to fully avoid that trap. There is a partial solution built in the script. In ecwam_runtime.sh, the variable ecwam_bathymetry_version=2 could be updated to 3, forcing the recalculation for this new release What do you think?

Yes, as Willem said, that's indeed the most robust way to do it, and every user will then be forced to recalculate the bathymetries once they update ecwam. Although I can't see that `create_wam_bathymetry_ETOPO1.F90` has been updated in this PR, is this related to another development? Just asking whether I should incorporate that when updating the norms.

src/programs/create_wam_bathymetry_ETOPO1.F90
jrbidlot, line 1: I have added a warning message.

The weird thing with this issue is that I had a very old set of 048 files that had been generated using grib1 rather than grib2. It must have been when I was testing/working on the grib output of those subgrib bathymetry files but I am not entirely sure.
So if want to be entirely sure, we might still want to update ecwam_bathymetry_version to 3 with this release.

I have not updated the norms in the 048 tests as I had computed them with the wrong files.
@awnawab has offered to recompute them anyway. Let me know if it is something I still need to do.

share/ecwam/scripts/ecwam_run_create_bathymetry.sh
```diff
@@ -173,12 +173,12 @@ else
     echo "\n\n\t Getting ETOPO1 data set\n"
     log ${SCRIPTS_DIR}/ecwam_retrieve.sh ${ETOPO1} ${DATA_DIR}/${ETOPO1}
     ln -sf ${DATA_DIR}/${ETOPO1} ETOPO1_Ice_g_int.xyz
-    CREATE_WAM_BATHYMETRY_EXE=${CREATE_WAM_BATHYMETRY_ETOPO1}
+    CREATE_WAM_BATHYMETRY_EXE=${CREATE_WAM_BATHYMETRY_ETOPO1}-${prec}
```
awnawab, line 176: I think ${prec} is undefined here, we would have to pass it as an argument e.g. `--prec=${prec}` in the call to `ecwam_run_create_bathymetry.sh` in `ecwam_run_preproc.sh`.

src/ecwam/sdice3.F90
```diff
@@ -127,8 +127,7 @@ SUBROUTINE SDICE3 (KIJS, KIJL, FL1, FLD, SL, SLICE,     &
 
            DO M = 1,NFRE
               DO IJ = KIJS,KIJL
-!                  ALP(IJ,M) = (2._JWRB*CDICE*(CITH(IJ)**(1.25_JWRB))*(FR(M)**(4.5_JWRB))) * ALPFAC(IJ) * ZALPFACB
-                 ALP(IJ,M) = (2._JWRB*CDICE*(CITH(IJ)**(1.25_JWRB))*(FR(M)**(4.5_JWRB))) * ALPFAC(IJ) ! (old way to ensure bit-identicality)
+                 ALP(IJ,M) = (2._JWRB*CDICE*(CITH(IJ)**(1.25_JWRB))*(FR(M)**(4.5_JWRB))) * ALPFAC(IJ) * ZALPFACB
```
awnawab, line 130: Does this mean the code is not bit-identical to the current state of 50R1?

awnawab (changes requested):
Thanks a lot @jrbidlot for bringing the testing up to date and making it scientifically more representative 🙏 All in all it looks good to me. I've left one small correction and have requested another clarification, once these are addressed we can merge this.

As for the testing, everything but some of the GPU tests and NVHPC SP tests are passing. I will fix these in a follow up PR once this PR is merged.

share/ecwam/scripts/ecwam_run_create_bathymetry.sh
jrbidlot: It does not seem to be necessary, but in any case I have added  your suggestion so that it is now clear how prec is entering ecwam_run_create_bathymetry

src/ecwam/sdice3.F90
jrbidlot: Thanks Ahmad for your careful review.

I have tested again, including in coupled mode with the IFS.
It is bit identical as far as I have tested. Note that currently ZALPFACB = 1
It might be why.
The reason I made the change is that we have now ZALPFACB in the namelist and a user might want to use a different value than 1. So we need the change.

awnawab (approved):
Thanks a lot @jrbidlot for the quick turnaround and explanations. This is now G2G ✅

src/ecwam/sdice3.F90
jkousal32: actually even with value ZALPFACB = 1 it was non-bit identical, that's why I had to put that line in. I think it was hitting non-bit identicality in CI testing.
So yes, even with ZALPFAC=1 it will be non bit identical wherever it gets pulled in to IFS

jkousal32 (approved):
it looks good to me (pending the note about bit-identicality), thanks alot for doing that work @jrbidlot .

src/ecwam/sdice3.F90
jrbidlot: We might want to do that test again, because it is not what I have found running a single single precision coupled forecast. I mean that one line change
@jkousal32 you decide what you want in that PR (as you are in charge of ecWAM), BUT be aware that from the point of view of an external user, such as MeteoFrance, the namelist and the documentation imply that one can change ZALPFAC but without that change nothing will happen.

If indeed it is not bit identical when coupled, we could always (temporarily) use the LWCOU flag  to only keep the old code when coupled. Be aware that the if statement should be outside of the do loops on M and IJ

jkousal32: I've just gone back through old messages with Zak. 

I extract the relevant information here:
experiments:

- injl: control v12 branches
- inye: non bit-identical branch. a set of passive changes going into 50r1. ecwam=1.5.0_wave_sea_ice
- inya: as above, but making it bit identical. ecwam=1.5.0_wave_sea_ice_bitIdentical. See commit https://github.com/jkousal32/ecwam/commit/cf2db3c3062e20183251b4088db65b453845c8b3
 
WAMNORMS were identical for injl & inya, whilst different for inye.

In commit https://github.com/jkousal32/ecwam/commit/cf2db3c3062e20183251b4088db65b453845c8b3 is the line in discussion here, but also lines 140-155 below (which currently remain the "old way"). Considering this surfaced in the wamnorms at the time, and it now seems to be bit identical as per your check @jrbidlot, probably the latter (lines 140-155 ) is the offending candidate, and I was overly cautious at the time (unless it is compiler differences between ecWAM and IFS?)

maybe we leave your commit here as it is Jean, i.e. remove this line.

thoughts @awnawab @jrbidlot ?

awnawab: I'll run a representative subset of the CI with Jean's branch, and if that passes then we can merge this.

jrbidlot: Thanks @awnawab

awnawab:
I've tested this branch in the fc_dp, fc_nemo4 and wam experiments run as part of the ifs staging CI, and it is bit-identical in all three, so I think we can be confident that this is indeed bitid to the current state of 50R1. Thanks all for the discussion 🙏

jkousal32:
> I've tested this branch in the fc_dp, fc_nemo4 and wam experiments run as part of the ifs staging CI, and it is bit-identical in all three, so I think we can be confident that this is indeed bitid to the current state of 50R1. Thanks all for the discussion 🙏

Great, thanks @awnawab !

PR #156, opened 2026-07-22
Fix GRIB2 streams for CY49R3 and incorporate recent changes from CY49R2

reuterbal:
### Description

This updates the standalone ecwam with some recent changes in CY49R2 and adjusts wave output for grib2 compliance.

### Contributor Declaration


src/ecwam/preset_wgrib_template.F90
jrbidlot, line 1: New line 173: MARSTYPE is used in CALL IGRIB_SET_VALUE(IGRIB_HANDLE,'type',MARSTYPE)
but MARSTYPE is character of length, whereas, I believe IGRIB_SET_VALUE requires a integer variable (?)

src/ecwam/wgribencode.F90
jrbidlot, line 1: Lines starting at line 383: WRITE(IU06,*) '   SUB. WGRIBENCODE: UNKNOWN MARS TYPE: '//MARSTYPE

wgribencode could be called from any mpi task and usually the logfile from task 1 is cat into the ecflow log file. I would therefore recommend writing the error message to NULERR instead (or as well)

src/ecwam/wstream_strg.F90
jrbidlot, line 1: It is probably too late now, but the reason for this routine was for when the wave and the atmosphere had different streams and one needed to find the corresponding stream (string and number) of one base on the other.
With the move the grib2 and the change to have the same stream, I would have assumed that this routine should no longer be needed

jrbidlot (commented):
I have made a few comments but otherwise it should be OK.

src/ecwam/preset_wgrib_template.F90
reuterbal: IGRIB_SET_VALUE is an interface that includes an IGRIB_SET_CHAR variant. From a type point of view this appears to be supported at least:

https://github.com/ecmwf-ifs/ecwam/blob/c9bec7b88f368367b251c3bd6be8ee3d05bf88fb/src/ecwam/yowgrib.F90#L48-L52

src/ecwam/wgribencode.F90
reuterbal: Thanks! I was copying the unit used in lines 392ff. Happy to change this!

src/ecwam/wstream_strg.F90
reuterbal: Quite possibly, but that's beyond my understanding. If this is to be changed I would suggest doing the same in CY49R2, too.

src/ecwam/preset_wgrib_template.F90
jrbidlot: Good to know.

src/ecwam/wstream_strg.F90
towil1: We still use wstream_strg.F90 to decide the wave stream when fetching pre-CY49R2 data from mars.

jrbidlot: Ok, then if it is the case, I don't understand the proposed changes.

I short, none of the changes should be applied.
or wstream_strg has wrongly been used for CY49R2 onwards

Details
The block with the change at line 529 changes KSTREAM from 1248 to 1247 which is also the value of the input ISTREAM. The ides is to return a different value for KSTREAM than ISTREAM. In that case, KSTREAM should stay = 1248 (the atmospheric stream number for atmospheric long cut-off)

The same for line 693, KSTREAM should stay 1250

For line 703:
CSTREAM should stay = ewla because that block returns a wave stream

awnawab: @reuterbal, @towil1 any further input? I'm not yet that familiar with grib and I/O, but if the primary use case for this routine is to retrieve pre 49R2 data, then as @jrbidlot suggests we shouldn't update this routine at all?

towil1: Yes, I'd agree with Jean and not change wstream_strg.F90, and also revert the changes in CY49R2. These edits go back a year ago by Cornel, and later updated by me in September.  Back then, it was still used for the setting of output wave stream, but these calls have since been removed.

reuterbal: Thanks for the feedback. I've reverted the changes to wstream_strg.F90

awnawab (approved):
Thanks all for the discussion and thanks @reuterbal for the quick turnaround 🙏 This is now G2G ✅

PR #157, opened 2026-07-31
Agent context and skills

awnawab:
This PR adds an AGENTS.md to provide extra context to agents when working with ecwam. A .fortls is also included, although this will only work if the fortls language server is on your path. It can help reduce token consumption significantly. Finally, a skill is also included that can be used to update the validation norms whenever there is a scientific or configuration change.


jkousal32 (approved):
Nice idea!

wdeconinck (approved):
Excellent!!! 👌

PR #159, opened 2026-08-11
fix for ifx flags and NSPEC2GRIB 

antonl321:
### Description

Re-add NSPEC2TAB=32 for CY49R3 arpifs multiom (needed by io_serv_multiom_utils_mod)

added flags for ifx (Intel-LLVM)

### Contributor Declaration


cmake/ecwam_compile_flags.cmake
```diff
@@ -38,6 +38,16 @@ elseif(CMAKE_Fortran_COMPILER_ID MATCHES "Intel")
   set(heap_arrays_flags   "-heap-arrays 32")
   set(optimization_flags  "-O2")
 
+elseif(CMAKE_Fortran_COMPILER_ID STREQUAL "IntelLLVM")
```
awnawab: Ah sorry about that, I should have cherry-picked ifx support to develop-1.4 too 😬 now that I look at it, the handling of ifx flags in develop is a bit lazy. Given that the only incompatible flag is `-fast-transcendentals`, we can just guard that one behind `CMAKE_Fortran_COMPILER_ID STREQUAL "Intel"` and keep the rest of the code unchanged. Or is that insufficient and I've missed something?

awnawab (changes requested):
Thanks a lot @antonl321 for these fixes 🙏 apologies that you had to add support for ifx flags, I should have cherry-picked the changes from develop. The implementation there in fact is a little messy, so we can try a neater solution here and I'll patch develop accordingly. (I have a pr open to add ifx to the CI in develop, I'll cherry-pick that here too once merged)

cmake/ecwam_compile_flags.cmake
antonl321: That's Claude verbosity :), useful if the two compiler flags keep diverging, hardly probable.
What about this solution used by Sam in ectrans
https://github.com/ecmwf-ifs/ectrans/commit/e9affb167351de572224c1e2b525ba3aea6f86a7

awnawab: That's precisely what I'm suggesting. `CMAKE_Fortran_COMPILER_ID MATCHES "Intel"` returns true for both ifort and ifx, so all we need is the `CMAKE_Fortran_COMPILER_ID STREQUAL "Intel"` guard around the specification of the fast transcendentals flags.

awnawab (approved):
Thanks a lot for the fix 🙏 merging now

PR #160, opened 2026-08-12
Restore CI and restrict loki to CI gpu builds

awnawab:
Previously, loki cpu builds were used for all the CI tests. This was implemented primarily for two reasons:
- In the absence of gpu testing, it was useful to have at least some coverage of the loki pipeline in the CI
- The loki cpu transformations were a very useful development aid

Both the above are no longer true, with the CI now able to run genuine GPU builds, and the loki integration and gpu functionality being fairly mature by this stage. We can therefore remove the loki cpu functionality. If there is a need to restore it to develop some cpu targeted optimisations, this can be restored via a single revert. In the absence of an actual use case, the maintenance effort for the loki cpu pipelines isn't worth it.

This will also allow the CI to become a more useful development aid, as the code running in (non-hpc) CI builds will be the same as the committed code. loki/gpu functionality will continue to be tested at the final PR stage, at which point I can help with debugging if necessary.

Alongside the above, I've also fixed some bugs highlighted during #154, including updating the tests that are run for debug builds.


jkousal32 (approved):

.github/workflows/build.yml
```diff
@@ -229,8 +229,8 @@ jobs:
...
           ecmwf-ifs/field_api: "-G Ninja -DCMAKE_BUILD_TYPE=${{ matrix.build_type }} -DENABLE_TESTS=OFF -DENABLE_ACC=OFF -DENABLE_SINGLE_PRECISION=${{ matrix.prec == 'SP' }} -DENABLE_DOUBLE_PRECISION=${{ matrix.prec == 'DP' }}"
           ecmwf-ifs/loki: "-G Ninja -DCMAKE_BUILD_TYPE=${{ matrix.build_type }} -DENABLE_TESTS=OFF"
-        cmake_options: "-G Ninja -DCMAKE_BUILD_TYPE=${{ matrix.build_type }} ${{ matrix.cmake_options }} -DENABLE_MPI=ON -DENABLE_LOKI=ON -DLOKI_MODE=idem-stack -DENABLE_SINGLE_PRECISION=${{ matrix.prec == 'SP' }} -DENABLE_DOUBLE_PRECISION=${{ matrix.prec == 'DP' }}"
-        ctest_options: "${{ matrix.ctest_options }} ${{ matrix.build_type == 'Debug' && '-R cy50r1' || '' }}"
+        cmake_options: "-G Ninja -DCMAKE_BUILD_TYPE=${{ matrix.build_type }} ${{ matrix.cmake_options }} -DENABLE_MPI=ON -DENABLE_SINGLE_PRECISION=${{ matrix.prec == 'SP' }} -DENABLE_DOUBLE_PRECISION=${{ matrix.prec == 'DP' }}"
+        ctest_options: "${{ matrix.ctest_options }} ${{ matrix.build_type == 'Debug' && '-R ^ecwam_test_etopo1_oper_an_fc_(O48)_(sp|dp)(_preproc|_preset|_mpi4_omp1|_mpi1_omp4)?$' || '' }}"
```
wdeconinck, line 244: Could you add a comment why these are filtered?

wdeconinck:
I see you added a new HPC arch. Would it make sense to add a CI for this build?

wdeconinck (approved):
Thanks for this @awnawab all good!

PR #161, opened 2026-08-31
Restore IFS nvhpc bitid

awnawab:
A small PR to restore bit-identicality in IFS coupled NVHPC builds. I've also replaced some manual flag handling with use of a new ecbuild macro. Finally, I've mopped up some leftover code related to loki cpu transformations which were deprecated in #160.


wdeconinck:
The ac-gpu nvhpc sp build is failing some unit tests, but output is not visible. Can we run the tests with  --output-on-failure?
And ideally find a fix :)

awnawab:
Ah, that's just because build-hpc is automatically picked up from main, which still has the problems I fixed in #160. Let me merge develop to main to pick up the fixed CI config.

jrbidlot:
Could the gpu failure also be connected to
https://github.com/ecmwf-ifs/ecwam/pull/163

wdeconinck (approved):
Thanks! LGTM

PR #163, opened 2026-08-31
Ecwam develop bugfix ctuw

jrbidlot:
### Description
There is bug in the OPENACC adaptation of ctuw.F90
Until recently, we did not have a test that would have triggered that part of the code because it needs to run with surface of currents. Such test now exist (etopo1_oper_an_fc_grib_restart_O96.yml)

I have also adapted to code so that the same action could not happen at different places depending on openacc or not

I have also added a comment that question why the block with the checks was remove from the openacc implementation

### Contributor Declaration


awnawab:
Hi @jrbidlot. Is this a duplicate of #162? If so, we should close one of them, and the right target of either would be develop not main.

jrbidlot:
I really don't understand how to use the github web interface.

I had requested only a copilot PR but then it went to you as well.

Let me try to clean this up

PR #164, opened 2026-08-31
Bugfix for OPENACC implementation and rationalisation of the code

jrbidlot:
### Contributor Declaration

### DIFF
--- a/src/ecwam/ctuw.F90
+++ b/src/ecwam/ctuw.F90
@@ -477,9 +477,8 @@ IF (LHOOK) CALL DR_HOOK('CTUW',0,ZHOOK_HANDLE)
               WKPMN(IJ,K,M,0)=(DTHP+ABS(DTHP))+(ABS(DTHM)-DTHM)
               WKPMN(IJ,K,M,1)=-DTHP+ABS(DTHP)
               WKPMN(IJ,K,M,-1)=DTHM+ABS(DTHM)
-#ifdef _OPENACC
+
               SUMWN(IJ,K,M)=SUMWN(IJ,K,M)+WKPMN(IJ,K,M,0)
-#endif
             ENDDO
           ENDDO
         ELSE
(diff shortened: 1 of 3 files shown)


jrbidlot:
Sorry, the github interface does not work for me.
So I ended up creating a new PR just for it to go to develop..

src/ecwam/ctuw.F90
```diff
@@ -533,6 +533,11 @@ SUBROUTINE CTUW (DELPRO, MSTART, MEND,                    &
 !!!   THE SUM IS NEEDED LATER ON !!!
 !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
 #ifndef _OPENACC
+
```
awnawab, line 536: We took a shortcut here because it would have been very expensive to copy the logical back to host and then trigger the exception. We now have a way of triggering the failure directly from the device, so no expensive copies back to host are needed.

After your PR is merged I'll port this path to gpu too and adapt it to trigger the failure directly from the device.

awnawab (approved):
Thanks for this bugfix @jrbidlot 🙏 This was indeed a genuine bug that you've fixed, but despite the bug the restart O96 test still passes as shown in #161.

And yes, we should restore the correctness check for gpu enabled builds too. I'll do so in a future PR as described in the inline comment.

jrbidlot:
@awnawab thank you for pointing that the bugfix was not actually picked up by the testing.

Well, it is because the O96 test did not run with surface currents, even though intended to run with them.
Simply because, I failed to add the necessary lines in the test yml file to point to the appropriate file.
I have now fixed file etopo1_oper_an_fc_grib_restart_O96.yml
Sorry about that.
I had a quick fix on the double precision norms. We still need to determine the single precision

awnawab:
Nice catch with spotting the bug in the test @jrbidlot! I hope you don't mind that I've pushed updated validation norms 😅

awnawab (approved):
Thanks a lot @jrbidlot for the prompt responses 🙏 with the fixes in place and the CI green again, this is now G2G ✅

PR #166, opened 2026-09-16
Remove unused CSTREAM variable

awnawab:
A small bugfix PR to remove the now unused CSTREAM variable from USERIN.


johannesbulin (approved):
Thanks for the fix! I guess this should be quite uncontroversial and should get the tests on LUMI running again.

awnawab:
We can ignore the failing gnu build-hpc test. This branch hasn't yet picked up the changes that enable compatibility with gnu 15, but PRs from forks use the workflow from the main branch, so it has still picked up the test.

src/ecwam/userin.F90
```diff
@@ -1237,7 +1236,6 @@ SUBROUTINE USERIN (IFORCA, LWCUR)
       WRITE(IU06,*) '  '
       WRITE(IU06,'("  CURRENT RUN:")')
       WRITE(IU06,'("  GRIB TABLE..: ", I4)') NLOCGRB
-      WRITE(IU06,'("  STREAM .....: ", A4)') CSTREAM
```
towil1, line 1240: In 50R2, I'm printing ISTREAM instead, see 5abede02c71d95f39e23d0b585b25866f0e12c20

awnawab: Thanks, that's a good idea, I've now updated my PR to do the same.

PR #168, opened 2026-09-21
AMD GPU offload

awnawab:
This PR adds support for offload to AMD GPUs, including a build-hpc CI entry for testing on lumi.

The approach followed is largely the same as that used for nvidia gpus, with the main difference that rather than openacc, the openmp target offload model has to be used. In its current state, this adds a lot of extra ifdef'ing in the code, especially in the wave propagation kernel under `PROPAG_WAM`.

This is not intended to stay in the code for long. Early next year, both sets of pragmas will be replaced with `!$loki parallel` pragmas, which at build time can then be translated to the target architecture of choice. We can then recover just one layer of ifdefing as before:
```
#ifdef WAM_GPU
!$loki parallel ....     ! gpu code path
#else 
!$OMP PARALLEL  ! cpu code path
#endif
```
 
Many thanks to @MichaelSt98 for helping develop this capability, and many thanks to @PaulMullowney for helping get this in a mature enough state to PR.


jkousal32 (approved):
Approved on the basis that it will be cleaned up next year; bringing it in now rather than then greatly helps maintainability. Nice work @awnawab !

src/ecwam/ctuwupdt.F90
MichaelSt98, line 206: shouldn't this be `#ifndef WAM_GPU`?

MichaelSt98 (approved):
Very nice!

src/ecwam/ctuwupdt.F90
awnawab: Yes! good catch

